PR #250, opened 2019-09-16
updates all version to a GM 2 version tag

chrisbsmith:

* With the advent of maintaining multiple version of our Grey Matter Helm charts, we quickly realized that keeping all published charts in the same `1.X.X` convention, we run the risk of have a fix for Grey Matter 1.0 accidentally overwrite a Grey Matter 2.0 Helm chart.  To prevent this, the Grey Matter 2.0 charts are now versioned as `2.X.X`.  When we release Grey Matter 3.0, we'll update the versions to `3.X.X`.  This will ensure that the charts meant for Grey Matter 2.0 won't be overridden by Grey Matter 1.0 updates.

None

Documented.

I've verified that the charts compile but have no manually deployed it yet.


gm-control-api/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
-appVersion: "0.2"
+appVersion: 0.8.1
```
tilleryd: good catch

tilleryd:
@chrisbsmith I think we should leave documentation as is in this PR. I'm not sure there's benefit to removing it yet. Also, there's a lot of config left over, if we do decide to remove it, that's enough to warrant another PR

tilleryd:
@chrisbsmith I ran this against staging "fabric-prod" successfully.

chrisbsmith:
@tilleryd  Agree with your comment on documentation. I've added the documentation folder back.

PR #252, opened 2019-09-18
change xds version and increment chart version

moosecanswim:
I updated the xds docker image tag and incremented the chart

### DIFF
--- a/xds/values.yaml
+++ b/xds/values.yaml
@@ -14,7 +14,7 @@ global:
 xds:
   image: 'docker.production.deciphernow.com/deciphernow/gm-xds:0.2.6'
   imagePullPolicy: Always
-  version: 0.2.6
+  version: 0.2.10
   cluster: greymatter
   #host: 'xds.greymatter.svc.cluster.local'
   port: 18000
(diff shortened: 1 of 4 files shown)


tilleryd:
@chrisbsmith Shouldn't this go into `master` like how we are merging PRs into `gm-2`? We should only merge into release branches as a side effect of merging into another long running branch, in this case master. We can cherry pick commits onto release branches...

moosecanswim:
https://jenkins.development.deciphernow.com/job/helm-chart/job/helm-charts/view/change-requests/job/PR-252/

moosecanswim:
This passed but it did not hit github

xds/values.yaml
```diff
@@ -14,7 +14,7 @@ global:
 xds:
   image: 'docker.production.deciphernow.com/deciphernow/gm-xds:0.2.6'
```
tilleryd: Shouldn't the Docker image tag be 0.2.10 as well?

tilleryd:
@chrisbsmith correct on merging into the release branches

chrisbsmith:
@moosecanswim this PR has also become `stale`.  And since we aren't using `xds` anymore, I don't think it's necessary to merge.  I will close this one by the end of the day unless I hear otherwise from you.

PR #287, opened 2019-10-02
fix YAML parsing errors

tilleryd:
@justincely and @joshua-rutherford, you both reported errors with YAML parsing. This should fix both cases.

Please re-deploy when you test this to confirm the fixes.

### DIFF
--- a/docs/Service Accounts.md
+++ b/docs/Service Accounts.md
@@ -30,7 +30,13 @@ The following list gives the service that needs a service account along with the
 - `spire-agent` - spire subchart - `.Values.spire.agent.serviceAccount`
 - `spire-server` - spire subchart - `.Values.spire.server.serviceAccount`
 
-If you're deploying into an environment where Helm/Tiller doesn't have sufficient permissions to create service accounts, you'll need to apply the [greymatter-service-accounts.yaml](../greymatter-service-accounts.yaml) file first. Be sure to set the appropriate namespace you're deploying to throughout that file. The user to apply this file will need to have sufficient permissions to create accounts. The custom values file will also need to be updated to ensure Helm doesn't try to create the accounts.
+If you're deploying into an environment where Helm/Tiller doesn't have sufficient permissions to create service accounts, you'll need to apply the [greymatter-service-accounts.yaml](../greymatter-service-accounts.yaml) file first. Be sure to set the appropriate namespace you're deploying to throughout that file. The user to apply this file will need to have sufficient permissions to create accounts.
+
+```sh
+oc apply -f greymatter-service-accounts.yaml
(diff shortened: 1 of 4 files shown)


greymatter.yaml
```diff
@@ -554,7 +554,6 @@ jwt:
         key: jwt.key.pem
 
   redis:
-    openshift:
```
joshua-rutherford: So this is only an issue on 2.14.0 not the documented 2.14.3 version of helm.  I don't know if we need to change this.

joshua-rutherford: But if we do we can change this in postgres too.

docs/Service Accounts.md
```diff
@@ -30,7 +30,13 @@ The following list gives the service that needs a service account along with the
...
 - `spire-server` - spire subchart - `.Values.spire.server.serviceAccount`
 
-If you're deploying into an environment where Helm/Tiller doesn't have sufficient permissions to create service accounts, you'll need to apply the [greymatter-service-accounts.yaml](../greymatter-service-accounts.yaml) file first. Be sure to set the appropriate namespace you're deploying to throughout that file. The user to apply this file will need to have sufficient permissions to create accounts. The custom values file will also need to be updated to ensure Helm doesn't try to create the accounts.
+If you're deploying into an environment where Helm/Tiller doesn't have sufficient permissions to create service accounts, you'll need to apply the [greymatter-service-accounts.yaml](../greymatter-service-accounts.yaml) file first. Be sure to set the appropriate namespace you're deploying to throughout that file. The user to apply this file will need to have sufficient permissions to create accounts.
+
+```sh
+oc apply -f greymatter-service-accounts.yaml
+```
+
+The custom values file will also need to be updated to ensure Helm doesn't try to create the accounts.
```
joshua-rutherford: This seems really vague.  Can we:

a. provide directions on what to remove.
b. provide a version of the custom config that does this.

greymatter.yaml
tilleryd: I'm ok with the change because now it's valid YAML.

docs/Service Accounts.md
tilleryd: Yes, I'll address this.

tilleryd: @joshua-rutherford Due to having 4 different custom files already in the repo I opted to clarify this all within this doc.

tilleryd:
@joshua-rutherford this is ready for another review

docs/Service Accounts.md
justincely: These instructions are going to evolve as we use it.  I'm ok with where it is now

PR #304, opened 2019-10-08
Adds the ability to enable basic observations per service

chrisbsmith:

This PR adds the ability to enable or disable observables per service.  

This new section is needs to be added to `.Values.global`:

```

<br/><br/>

<br/><br/>

### DIFF
--- a/docs/Getting Started.md
+++ b/docs/Getting Started.md
@@ -62,6 +62,13 @@ global:
 
   # whether to include the virtual cluster namespace in the final hostname of the deployment
   remove_namespace_from_url: false
+
+  # a map of information specific for Grey Matter observables
+  observables:
+    # the Kafka topic that the observables should be written to
+    topic: observables
+    # the Kafka server connection string
+    kafkaServerConnection: 
 ```
 
 ### Ingress
@@ -94,7 +101,7 @@ Read [Ingress](./Ingress.md) for further details.
 
 ### Observables
 
-The Grey Matter Helm chart provides the ability to enable observables in the mesh, but they are disabled by default.  To enable observables, modify the following settings in your local custom values file.  The `kafkaServerConnection` should be a reference to a kafka ensemble that is available in the mesh.  The format is `<host1>:<port1>,<host2>:<port2>`
+The Grey Matter Helm chart provides the ability to enable observables in the mesh, but they are disabled by default.  To enable observables, modify the following settings in your local custom values file.  The `kafkaServerConnection` should be a reference to a Kafka ensemble that is available in the mesh.  The format is `<host1>:<port1>,<host2>:<port2>`
 
 These are global settings for all observables:
 
@@ -105,7 +112,7 @@ globals:
     kafkaServerConnection: 
 ```
 
-Observables can be enabled or disabled for each service.  You can enable observables by setting `.Values.global.services.<service>.observables` to `true` or `false`
+Observables can be enabled or disabled for each service.  You can enable observables by setting `.Values.global.services.<service>.observablesEnabled` to `true` or `false`
 
 ### Docker credentials
 
--- a/docs/Getting Started.md
+++ b/docs/Getting Started.md
@@ -5,6 +5,7 @@
   - [OpenShift](#openshift)
   - [Configuration](#configuration)
     - [Ingress](#ingress)
+    - [Observables](#observables)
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: docs/Getting Started.md [hunk 1]
```diff
@@ -91,6 +92,21 @@ Once the edge proxy is deployed, voyager-operator will create a custom ingress r
...
+
+The Grey Matter Helm chart provides the ability to enable observables in the mesh, but they are disabled by default.  To enable observables, modify the following settings in your local custom values file.  The `kafkaServerConnection` should be a reference to a kafka ensemble that is available in the mesh.  The format is `<host1>:<port1>,<host2>:<port2>`
+
+These are global settings for all observables:
+
+```yaml
+globals:
+  observables:
+    topic: observables
+    kafkaServerConnection:
```
tilleryd: I think we should provide valid YAML. A key without a value throws errors with older versions of Helm (despite we don't support them). Also it's just good practice.

greymatter/values.yaml
```diff
@@ -4,6 +4,12 @@ global:
   route_url_name:
   remove_namespace_from_url: false
   zone: zone-default-zone
+  # The observables section defines the common information points for
+  # Grey Matter Observables.  Note that, if enabled, the kafkaServerConnection
+  # MUST be enabled
+  observables:
+    topic: observables
+    kafkaServerConnection:
```
tilleryd: Again, I think this key should have a value.

docs/Getting Started.md
chrisbsmith: So I thought about this, but if I provide something like `localhost:9092` and the user enables it without changing this value, there are more severe consequences in the mesh.  Open to it if you really thing we should.

greymatter/values.yaml
```diff
@@ -68,6 +74,7 @@ global:
       catalogEnabled: true
       controlEnabled: true
       serviceName: 'catalog'
+      observables: false
```
tilleryd: We should be consistent with boolean naming conventions. e.g. `observablesEnabled`.

gm-control-api/json/services/proxy.json
```diff
@@ -9,9 +9,23 @@
...
+    {{-  if (.service.observables) }}
+    "gm_observables": {
+      "useKafka": true,
+      "topic": "{{.service.serviceName}}",
+      "eventTopic": "{{ .Values.global.observables.topic }}",
+      "kafkaServerConnection": "{{ .Values.global.observables.kafkaServerConnection }}"
+    }
+    {{- end }}
     "gm_metrics": {
+      "service_observable": "{{.service.observables}}"
```
tilleryd: I'm not aware of this as a k/v pair used by metrics. What does it do?

greymatter/values.yaml
chrisbsmith: 👍

docs/Getting Started.md
```diff
@@ -91,6 +92,21 @@ Once the edge proxy is deployed, voyager-operator will create a custom ingress r
 
 Read [Ingress](./Ingress.md) for further details.
 
+### Observables
+
+The Grey Matter Helm chart provides the ability to enable observables in the mesh, but they are disabled by default.  To enable observables, modify the following settings in your local custom values file.  The `kafkaServerConnection` should be a reference to a kafka ensemble that is available in the mesh.  The format is `<host1>:<port1>,<host2>:<port2>`
```
tilleryd: `Kafka` (uppercase)

tilleryd: I updated this doc to instruct the user to create a `custom-greymatter.yaml` file. I suggest we have them add this config there. i.e. `modify the following settings in custom-greymatter.yaml.`.

chrisbsmith: done.

gm-control-api/json/services/proxy.json
chrisbsmith: Good catch, that was cruft I forgot to remove.

docs/Getting Started.md
tilleryd: Hmm, now that I think about it, there isn't a value to add until they've setup Kafka. So I guess this is adequate.

greymatter/values.yaml
tilleryd: Never mind, same realization as my comment on the Getting Started doc.

PR #322, opened 2019-10-16
Add default consul values to the slo chart

### LINKED ISSUE #321: slo missing global.consul.enabled check
The SLO chart is missing the `.Values.global.consul.enabled` in slo/values.yaml.  This is causing templating errors

chrisbsmith:

Resolves #321 

The SLO chart did not have a default setting for consul, so if someone tried to install Grey Matter with an old yaml file, it would fail.

n/a

n/a


tilleryd:
@chrisbsmith Is it our strategy to always add global variables to each chart? I can see us bloating all of the charts in this manner. Wouldn't a better strategy to always point users to the latest ./greymatter.yaml file?

chrisbsmith:
@tilleryd global variables only need to be added to the chart that requires them.  For instance, in this example, the new global variable `.Values.global.consul.enabled` was added to slo.  When Parth attempted to deploy the Grey Matter chart without the updates to `greymatter.yaml`, SLO failed.  Each chart should be able to be deployed individually (even though the Grey Matter Helm chart), so each chart's value.yaml file should contain all variables it needs to run.

If other charts don't need `.Values.global.consul.enabled` then there's no need to create that option to the chart's values.yaml file.

tilleryd:
@chrisbsmith Got it. @zoemccormick I suggest just adding to the PR for the other charts.

tilleryd:
@chrisbsmith I just successfully installed without providing any global Consul configs in `./custom-greymatter.yaml`. Once the conflicts are resolved and chart versions are updated I'll approve.

PR #342, opened 2019-10-22
Update/add release to data names

moosecanswim:

updates the data chart so that a user can specify not to create resources when installing.   This PR also adds the release name to the name of resources created to allow for multiple deployments of data in a single namespace.

No changes to custom.yaml file when deploying the full greymatter suite.  when deploying standalone data you can use these values to avoid the additional creation of undesired resources.
```
    # To specify a resource name insteard of creating one set create to false and define a manual_name
    # During a standalone deployment some resources may be present in the namespace already.  In that case
    # you can avoid creating the resource (create: false) and instead point the chart to that resource 
    # with `manual_name`. (mongo_configmap.manual_name is not used because the password secret is passed
    # in via envar and internally defined as mongo-credentials)
    resources:
      data_secret:
        create: true
        manual_name: ""
      docker_secret:
        create: true
        manual_name: ""
      mongo_certs:
        create: true
        manual_name: ""
      mongo_configmap:
        create: true
        manual_name: ""
      mongo_passwords:
        create: true
        # manual_name: "mongo-credentials"
      jwt_secret:
        create: false
        manual_name: ""
```

<br/><br/>

Jenkins passes.  I've deployed the full gm suite as well as a standalone into the development cluster successfully.

### DIFF
--- a/data/Chart.yaml
+++ b/data/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 0.2.3
 description: A Helm chart to deploy Grey Matter Data
 name: data
-version: 1.0.5
+version: 1.0.6
 icon: https://s3.amazonaws.com/decipher-public/grey-matter/branding/grey-matter-mark-rgb.png
 maintainers:
   - name: Decipher Technology Studios Engineering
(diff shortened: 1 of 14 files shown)


moosecanswim:
stale and addressed in in gm2- implementation

PR #358, opened 2019-10-24
docs: using helm to generate k8s config

### LINKED ISSUE #353: Create doc for installing on raw k8s without Helm/Tiller
Recently, we provided raw k8s YAML generated from a dry-run of our Helm charts. This is a quick way to get a customer up and running on k8s if they can't use Helm/Tiller. @kaitmore has a nice templated solution on top of this to apply values similar to what we are doing with Helm.

Let's get this documented as a fallback to Helm in this repo.

kaitmore:

<br/>
Resolves #353 
<br/>

<br/>n/a<br/>

<br/>n/a<br/>

<br/>n/a<br/>

### DIFF
--- a/docs/Generating k8s Config.md
+++ b/docs/Generating k8s Config.md
@@ -1,0 +1,61 @@
+# Generating k8s Configuration with Helm
+
+It is sometimes necessary to deploy Grey Matter without the use of Helm/Tiller. Luckily, Helm provides commands that can be used to generate raw kubernetes configuration.
+
+## Configuring the deployment
+
+If you know all the custom values necessary to generate the deployment ahead of time, fill out `greymatter.yaml` and `greymatter-secrets.yaml` which are hosted in the [Decipher Helm charts repo](https://github.com/DecipherNow/helm-charts) and [run the template command](#generating-templated-k8s-config) with them.
+
+If deployment configuration is not known beforehand, for example if the template is being passed off to a client, you can just add placeholder values to `greymatter.yaml`.
+
+## Templating
+
+The most straightforward way to generate configuration is to clone the [Decipher Helm charts repository](https://github.com/DecipherNow/helm-charts) and run the the following command:
+
+```sh
+helm template greymatter -f greymatter.yaml -f greymatter-secrets.yaml
+```
+
+The template command can also be run on charts hosted in remote repositories, albeit not directly. To get around that, first fetch and untar the remote charts before running the templating command:
+
+```sh
+helm fetch --untar --untardir . decipher/greymatter
+helm template greymatter -f greymatter.yaml -f greymatter-secrets.yaml
+```
+
+### Docker Images
+
+The following images are expected to be available in the configured docker repository:
+
+- `gm-catalog:1.0.1`
+- `gm-proxy:0.9.1`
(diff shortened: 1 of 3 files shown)


docs/Generating k8s Config.md
```diff
@@ -0,0 +1,42 @@
+# Generating k8s Configuration with Helm
+
+It is sometimes necessary to deploy Grey Matter without the use of Helm/Tiller. Luckily, Helm provides commands that can be used to generate raw kubernetes configuration.
```
chrisbsmith: I would say `without the use of Tiller`.  Helm is still used in this equation.

docs/Generating k8s Config.md
```diff
@@ -0,0 +1,42 @@
...
+The most straightforward way to generate configuration is to clone the [Decipher Helm charts repository](https://github.com/DecipherNow/helm-charts) and run the the following command:
+
+```sh
+helm template greymatter -f greymatter.yaml -f greymatter-secrets.yaml > template.yaml
+```
+
+The template command can also be run on charts hosted in remote repositories, albeit not directly. To get around that, first fetch and untar the remote charts before running the templating command:
+
+```sh
+helm fetch --untar --untardir . decipher/greymatter
```
chrisbsmith: Do we need to add a line about this about adding the decipher helm repo?  Are these instructions expected to be a complete guide or a supplement to go along with the others?

kaitmore:
@chrisbsmith Thanks for the comments! I've changed the file name and added more prereqs/links to the minikube doc. I figure this doc will be more of a supplement.

docs/Generating k8s Config.md
kaitmore: I linked to other documentation 👍

kaitmore:
@tilleryd 

`helm template` runs through all charts and templates out every file. After looking through helm docs I don't see any way to tell it not to template out specific files, tests or otherwise. 

I could either put a note in the doc saying to delete connection tests if not desired, or I could add a variable to each test yaml, such as `{{- if .gm-control-api.enableTest }}`, although that might be out of scope for this PR.

tilleryd:
I think we should add a variable to turn off the tests. They haven't been a factor since we've been deploying the charts. @chrisbsmith I'm curious on your thoughts here.

kaitmore:
@tilleryd My only other hesitation is that "turning off tests" is a bit misleading. Running `helm install` doesn't run tests - so setting the `enableTests` value wouldn't do anything in that scenario, it would only be used for this specific case where we're templating.

chrisbsmith:
I will say that we took an initial dive into using the Helm tests as an opportunity to see if it would be useful for us.  It may be in the long term, but we have not done anything to keep these files up to date.  You can see that it's trying to find catalog at `/services/catalog/1.2`.  I'd be OK removing the files all together but possibly doing more research into leveraging the `helm test` feature.  

@danielpcox  may be interested in this as well

tilleryd:
I say we delete them since gm-integrity will satisfy the mesh verification tests.

chrisbsmith:
no issues here, but I do think the test hook may be a useful feature to deploy gm-integrity into k8s automatically.

Issue #263, opened 2019-09-20
Improve documentation by adding descriptions to values


tilleryd:
We need to add descriptions to all of our value files so `gen-docs.py` populates the descriptions in our auto generated chart documentation.

tilleryd:
Conversation from #368 ...

<img width="631" alt="Screen Shot 2019-10-30 at 3 33 48 PM" src="https://user-images.githubusercontent.com/2581080/67892273-c2635a00-fb2a-11e9-9767-e9d3b9d02f6d.png">

tilleryd:
It appears that we now have comments above keys in values.yaml files throughout the charts. We should look at `gen-docs.py` and basically hook up the descriptions to the comments.

PR #368, opened 2019-10-29
Dynamically provision the prometheus namespaces to monitor

### LINKED ISSUE #363: Prometheus recording rules need to match the namespaces Control is monitoring
As currently written, Prometheus will only ever scrape the Grey Matter namespace for metrics:
https://github.com/DecipherNow/helm-charts/blob/c71db206edf22ec203533f9fba15ed4fd286f8d6/dashboard/templates/prometheus-configmap.yaml#L43

The charts need to be able to scrape metrics in this namespace as well as any other that Control is monitoring.

https://github.com/DecipherNow/helm-charts/blob/c71db206edf22ec203533f9fba15ed4fd286f8d6/control/values.yaml#L19

This will only resolve any issues with the recording rules when Helm installs Grey Matter, and could actually be potentially overwritten if they are changed on disk and a Helm update is run later.  

Documentation needs to be updated to make sure this update is captured.

chrisbsmith:

The prometheus-configmap was only ever generated to monitor the namespace where Grey Matter was deployed.  GM Control is configured to monitor multiple namespaces, but if Prometheus never was, then the Dashboard metrics would never work.

The helper template was updated to loop over the same values used by Control

resolves #363 

What was `.Values.control.additionalNamespacesToControl` has been moved to `.Values.global.control.additionalNamespacesToControl` so the dashboard chart could see it as well.

n/a

*in progress, will remove WIP once complete*


control/values.yaml
```diff
@@ -8,6 +8,7 @@ global:
     replicas: 1
   control:
     version: 1.0.2
+    additionalNamespacesToControl:
```
tilleryd: the control chart documentation should be updated with an example of what the value should be i.e. is it an array of string or a comma delimited string?

tilleryd: i see in the dashboard _helpers.tpl that it's an array of strings but it's not clear without looking at how its used

dashboard/templates/_helpers.tpl
```diff
@@ -7,4 +7,23 @@ Define the exhibitor host.
...
+*/}}
+{{- define "greymatter.dashboard.prometheus_namespaces" -}}
+{{- $namespaces := dict "namespaces" (list) -}}
+{{- $noop := printf "%s" $.Release.Namespace | append $namespaces.namespaces | set $namespaces "namespaces" -}}
+{{- if $.Values.global.control.additionalNamespacesToControl -}}
+{{- range $ns, $e := splitList "," $.Values.global.control.additionalNamespacesToControl -}}
+{{- $noop := printf "%s" $e | append $namespaces.namespaces | set $namespaces "namespaces" -}}
+{{- end -}}
+{{- end -}}
+{{/*}} {{- join "\n" $namespaces.namespaces | quote -}} */}}
```
tilleryd: this seems like dead code

tilleryd:
@chrisbsmith Did you tear down the deployment? I only see DHS Test namespace in staging.

chrisbsmith:
@tilleryd I did.

dashboard/templates/_helpers.tpl
chrisbsmith: Good catch, removing.

control/values.yaml
chrisbsmith: Agreed, I updated the configuration.md

control/configuration.md
```diff
@@ -1,32 +1,32 @@
...
-|global.waiter.image             |           |deciphernow/k8s-waiter:latest|
+|                 Parameter                  |Description|           Default           |
+|--------------------------------------------|-----------|-----------------------------|
+|global.environment                          |           |kubernetes                   |
+|global.domain                               |           |                             |
+|global.route_url_name                       |           |                             |
+|global.remove_namespace_from_url            |           |                             |
+|global.exhibitor.replicas                   |           |                            1|
+|global.control.version                      |           |1.0.2                        |
+|global.control.additionalNamespacesToControl|           |                             |
```
tilleryd: ah, i guess we can't supply a default because we never really know the namespace we're deploying to

tilleryd: and we still haven't resolved the lack of descriptions, which would help here too...

chrisbsmith: Yeah, I just ran the `gen_docs.py` script, which doesn't capture any descriptions.  It would probably be worth having someone update this script to collect descriptions out of the code, and then we could run it as part of the commit webhook since most people will probably forget.

tilleryd: Yeah https://github.com/DecipherNow/helm-charts/issues/263

PR #372, opened 2019-10-31
Voyager Minikube

whenhellfreezes:
…patches and then installs.

Voyager's latest release has a small problem. This commit updates voyager to latest and applies a patch before installing. Updating to the latest voyager will let us move our k8s version to 1.16 which comes paired with minikube 1.4.0. That means that minikube 1.3.1 will be depreciated for helm-chart support soon.

<br/><br/>

NA
<br/><br/>

No. This is something to discuss though. How would we document this? 

<br/><br/>

Ran from scratch (minikube 1.4.0) and pulled up the dashboard successfully. All pods running.
<br/><br/>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.tgz
 custom*.yaml
 credentials.yaml
+ci/scripts/manifest.yaml
 greymatter/charts
 charts.txt
 *.log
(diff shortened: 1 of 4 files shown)


whenhellfreezes:
This probably should go with some documentation changes as well.

whenhellfreezes:
So the `Deploy with Minikube` docs already say minikube@1.4.0 which is part of why I made these fixes, because before this commit that wasn't quite correct.

Issue #376, opened 2019-11-04
Investigate catalog-init script


kaitmore:
Catalog will occassionally start up without all the configured services:

![image](https://user-images.githubusercontent.com/22768918/68127967-e507c100-fee4-11e9-8896-d3f376e9ceb3.png)

tilleryd:
Need to dig into the catalog-init job https://github.com/DecipherNow/helm-charts/blob/release-2.1/catalog/templates/catalog-init.yaml

chrisbsmith:
This _may_ be an issue caused by the lag of catalog persisting data to GM Data.  The `curl` commands that issue the init job are kind of fire and forget.  They don't retry, rather they fail and the entire script should exit.

The actual script is located here: https://github.com/DecipherNow/helm-charts/blob/release-2.1/catalog/config/bootstrap.sh

kaitmore:
@chrisbsmith @tilleryd @danielpcox 

After investigating and trying various fixes, I've concluded that the catalog-init script is not the issue here. My first approach was to modify the script to retry until the request succeeds, however, I still occasionally saw missing services in the dashboard. 

In the following case it was the JWT service. You can see the successful 200 response from catalog here after running the catalog-init job:

```
Found service: jwt-security/
< HTTP/1.1 200 OK
< Content-Type: application/json
< X-Request-Id: 
< Date: Fri, 22 Nov 2019 21:21:51 GMT
< Content-Length: 27
< 
{ [27 bytes data]

100   473  100    27  100   446   1588  26235 --:--:-- --:--:-- --:--:-- 27823
* Connection #0 to host catalog left intact
Creating catalog item service.json
Contents of service.json...

{
  "clusterName": "jwt-security",
  "zoneName": "zone-default-zone",
  "name": "Grey Matter JWT Security",
  "version": "0.2.0",
  "owner": "Decipher",
  "capability": "Grey Matter",
  "runtime": "GO",
  "documentation": "",
  "prometheusJob": "",
  "minInstances": 1,
  "maxInstances": 1,
  "authorized": true,
  "clusterID": "",
  "meshID": "",
  "enableInstanceMetrics": true,
  "enableHistoricalMetrics": true,
  "metricsTemplate": "",
  "metricsPort": 8081
}
Uploading service.json...

* TCP_NODELAY set
* Connected to catalog (10.104.127.197) port 9080 (#0)
* ALPN, offering http/1.1
* successfully set certificate verify locations:
[...]

PR #379, opened 2019-11-05
378 storageclass

### LINKED ISSUE #167: Any OpenShift Service Resources should be a of type ClusterIp with an IP assigned
Based on information [this article](https://medium.com/faun/kubernetes-headless-service-vs-clusterip-and-traffic-distribution-904b058f0dfd) , OpenShift service resources should not declare `clusterIP: None`.  This creates a service resource that does DNS look up, and not load balancing. Leaving 'cluserIP: None` off allows the resource to load balance requests

### LINKED ISSUE #378: StatefulSets need to be able to pass a StorageClassName
The mongo charts currently don't provide the option to provide a storageClassName for the PersistentVolumeTemplate.  In some deployments, this causes an issue and the mongo pods will not deploy.

Consul provides a good example of how to set this.  https://github.com/helm/charts/blob/d4a1484cfeb35e9e36ea40c5bda2cedf91107b22/stable/consul/templates/consul-statefulset.yaml#L263-L278

@tilleryd

chrisbsmith:

* Remove the `ClusterIP: None` from any service.  This prevents k8s from performing proper load balancing to an service
* Allows a user to define the `StorageClassName` for the mongo PersistentVolumeTemplate.

resolves #167 
resolves #378 

Mongo PVC info was moved into a new grouping:
```yaml
mongo:
  storage:
    size: 40
    #storageClassName: "-"
```

For minikube, EKS and OpenShift, there is no change.  Some other Kubernetes deployments may need to set the storageClass Name

_testing in progress_

### DIFF
--- a/docs/Getting Started.md
+++ b/docs/Getting Started.md
@@ -1,25 +1,24 @@
 # Getting Started
 
-- [Getting Started](#getting-started)
-  - [Helm](#helm)
-  - [OpenShift](#openshift)
-  - [Configuration](#configuration)
-    - [Storage](#storage)
-    - [Ingress](#ingress)
-    - [Observables](#observables)
-    - [Docker credentials](#docker-credentials)
-    - [AWS credentials (optional)](#aws-credentials-optional)
-    - [Certificates](#certificates)
-    - [SPIRE](#spire)
-    - [TLS Options](#tls-options)
-    - [Single-service deployments](#single-service-deployments)
-  - [Install](#install)
-    - [Prepare Tiller](#prepare-tiller)
-    - [Prepare Service Accounts](#prepare-service-accounts)
-    - [Latest Helm charts release](#latest-helm-charts-release)
-    - [Local Helm charts](#local-helm-charts)
-    - [Additional Helm install flags](#additional-helm-install-flags)
-  - [Verification](#verification)
+- [Helm](#helm)
+- [OpenShift](#openshift)
+- [Configuration](#configuration)
+  - [Storage](#storage)
+  - [Ingress](#ingress)
+  - [Observables](#observables)
+  - [Docker credentials](#docker-credentials)
+  - [AWS credentials (optional)](#aws-credentials-optional)
+  - [Certificates](#certificates)
+  - [SPIRE](#spire)
+  - [TLS Options](#tls-options)
+  - [Single-service deployments](#single-service-deployments)
+- [Install](#install)
+  - [Prepare Tiller](#prepare-tiller)
+  - [Prepare Service Accounts](#prepare-service-accounts)
+  - [Latest Helm charts release](#latest-helm-charts-release)
+  - [Local Helm charts](#local-helm-charts)
+  - [Additional Helm install flags](#additional-helm-install-flags)
+- [Verification](#verification)
 
 This guide assumes that your target environment is a hosted Kubernetes based platform. If you want to test drive Grey Matter on your local machine, follow [Deploy with Minikube](./Deploy%20with%20Minikube.md).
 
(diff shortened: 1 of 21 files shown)


docs/Getting Started.md
```diff
@@ -70,6 +72,10 @@ global:
     kafkaServerConnection:
 ```
 
+### Storage
+
+The Grey Matter Helm chart assumes that your Kubernetes cluster has a default storage provider already defined.  The charts will attempt to create several PersistentVolumeClaims for data storage, and if a default storage provider has not been declared, the installation will fail.  Additionally, the Mongo StatefulSets declare a Persistent Volume Template has that the option to provide a different Storage Class Name.  This variable can be set at `.Values.data.mongo.storage.storageClass` and `.Values.internal-data.mongo.storage.storageName`
```
tilleryd:
```suggestion
The Grey Matter Helm chart assumes that your Kubernetes cluster has a default storage provider already defined.  The charts will attempt to create several PersistentVolumeClaims for data storage, and if a default storage provider has not been declared, the installation will fail.  Additionally, the Mongo StatefulSets declare a Persistent Volume Template that has the option to provide a different Storage Class Name.  This variable can be set at `.Values.data.mongo.storage.storageClass` and `.Values.internal-data.mongo.storage.storageName`
```

Since we don't define a a default storage class I assume the problem only occurs when the default storage class doesn't exist and an existing name hasn't been applied in it's replacement.

docs/Getting Started.md
```diff
@@ -1,23 +1,25 @@
...
-  - [TLS Options](#tls-options)
-  - [Single-service deployments](#single-service-deployments)
-- [Install](#install)
-  - [Prepare Tiller](#prepare-tiller)
-  - [Prepare Service Accounts](#prepare-service-accounts)
-  - [Latest Helm charts release](#latest-helm-charts-release)
-  - [Local Helm charts](#local-helm-charts)
-  - [Additional Helm install flags](#additional-helm-install-flags)
-- [Verification](#verification)
+- [Getting Started](#getting-started)
```
tilleryd: Are you using Markdown All In One for VSCode? I have it configured to ignore the top level header.

Screen Shot 2019-11-05 at 5 27 04 PM

tilleryd:
@chrisbsmith Can you provide YAML for setting up a StorageClass so I can test this for real?

docs/Getting Started.md
chrisbsmith: I do and I updated mine to match your settings.  We probably need to publish this to all developers though.

chrisbsmith:
@tilleryd I don't know what the actual `StorageClassName` should be, but you would supply it below.  You can add some value and see that it does in fact tempalte.

```
data:
  mongo:
    ssl:
    storage:
      storageClass: 'someStorageClass'
```

docs/Getting Started.md
chrisbsmith: Wow, that was a total dyslexia mistake, thanks for the catch.  I think I addresses your comment with my update.

chrisbsmith:
@tilleryd I also noticed that we never bumped all of the charts to `2.1.X` when we moved to the `release-2.1` branch.  I did this in the last commit, so if we want to back it out and apply it separately we can, but I thought this would suffice.

chrisbsmith:
Testing complete.  Performed a local deploy to minikube with success.  However, I did notice that the issue @kaitmore documented (#376) struck again.  GM Data did not appear in the dashboard, but viewing the logs from catalog-init, catalog returned a `200` when the data catalog.json was applied.

docs/Getting Started.md
tilleryd: I'll send an email to the company.

tilleryd:
@chrisbsmith I was asking about setting up a StorageClass like below. Did you do that? Obviously, not using aws-ebs but something that works with Minikube.

```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: standard
provisioner: kubernetes.io/aws-ebs
parameters:
  type: gp2
reclaimPolicy: Retain
allowVolumeExpansion: true
mountOptions:
  - debug
volumeBindingMode: Immediate
```

chrisbsmith:
@tilleryd  I did not.  I'm not sure exactly how to do that in a minikube environment; specifically what a StorageClass would be on a local installation.  We can attempt to set up kube-spray and portworx like they had if you want to dig that deep.

PR #380, opened 2019-11-06
enable mtls for control-api

kaitmore:

Enable TLS for gm-control-api
<br/>
<br/>

<br/>
<br/>

<br/><br/>

To test, assuming you have minikube running:

- update requirements.yaml to point to the local charts for control and gm-control-api
- install greymatter:

 ```
helm dep up greymatter
helm install greymatter -f greymatter.yaml -f greymatter-secrets.yaml -f credentials.yaml --set global.environment=kubernetes -n gm-deploy
```
- look at logs for gm-control-api and see that tls was enabled
- check you can reach the gm-control-api through the edge
- delete the deployment
- in greymatter.yaml, set global.control_api_tls=false and global.services.controlApi.egressTLS=false
- repeat previous steps with tls false to make sure everything is working
<br/>

### DIFF
--- a/control/values.yaml
+++ b/control/values.yaml
@@ -3,6 +3,7 @@ global:
   environment: kubernetes
   domain:
   route_url_name:
+  control_api_tls: true
   remove_namespace_from_url: ''
   exhibitor:
     replicas: 1
@@ -21,6 +22,23 @@ control:
   replicas: 1
   image: 'docker.production.deciphernow.com/deciphernow/gm-control:{{ $.Values.global.control.version }}'
   imagePullPolicy: IfNotPresent
+  controlApiTLS:
+    name: control-to-control-api-certs
+    mountPoint: /service-certs
+    insecure: true
+    certificates:
+      ca: |-
+        -----BEGIN CERTIFICATE-----
+        ...
+        -----END CERTIFICATE-----
+      cert: |-
+        -----BEGIN CERTIFICATE-----
+        ...
+        -----END CERTIFICATE-----
+      key: |-
+        -----BEGIN RSA PRIVATE KEY-----
+        ...
+        -----END RSA PRIVATE KEY-----
   serviceAccount:
     create: true
     name: control-sa
@@ -34,9 +52,6 @@ control:
     gm_control_api_zone_name:
       type: 'value'
       value: 'default-zone'
-    gm_control_api_ssl:
-      type: 'value'
-      value: 'false'
     gm_control_api_host:
       type: 'value'
       value: 'gm-control-api:5555'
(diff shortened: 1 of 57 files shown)

### CODE REVIEW

**Comment by kaitmore** [comment]
@dgoldstein1 Thanks! You're right, charts should be bumped.

I took the helm install command from the Makefile. That was easiest for me since I already generated the `credentials.yaml` file by running `make credentials`

**Comment by tilleryd** [comment]
I also run the Makefile style install because it lessens copying or changing greymatter.yaml or greymatter-secrets.yaml. Having only credentials.yaml and the flag `--set global.environment=kubernetes` is all we need.

File: control/values.yaml [hunk 1]
```diff
@@ -3,6 +3,7 @@ global:
   environment: kubernetes
   domain:
   route_url_name:
+  control_api_tls: true
```
tilleryd: This isn't a global config to me and seems more appropriate under `gm-control-api`.

tilleryd: I'm aware that you will need to change a bunch of files to address my concern, so I'm open to a debate.

gm-control-api/templates/gm-control-api.yaml
```diff
@@ -75,6 +90,29 @@ spec:
...
+          # We give a full minute before we start health-checking for failures
+          livenessProbe:
+            exec:
+              command:
+                - /bin/sh
+                - -c
+                - "curl -k --cacert /etc/proxy/tls/sidecar/ca.crt --key /etc/proxy/tls/sidecar/server.key --cert /etc/proxy/tls/sidecar/server.crt https://localhost:5555/v1.0/cluster"
+            initialDelaySeconds: 60
+            periodSeconds: 3
+          {{- end }}
```
tilleryd: Nice addition

tilleryd:
I tested TLS and non-TLS, and it worked as intended. My only comment is on the global value.

gm-control-api/templates/gm-control-api.yaml
chrisbsmith: So I admittedly don't know enough about `livenessProbes` and `readinessProbes`, but was just reading this article today (https://learnk8s.io/production-best-practices/) and it recommends not using `livenessProbes` if possible.  And recommends the two probes be different.

This is really more food for thought, but we should definitely start including readiness probes in all our containers, and making sure they fail properly so k8s can restart them.

control/values.yaml
kaitmore: @tilleryd This was the solution I discussed with @chrisbsmith. I agree it's not ideal, but the issue is that both control and control-api need to read this variable, and control wouldn't be able to if I kept it under gm-control-api as you suggested.

Open to other suggestions, it seemed messy to me to have to enable control-api tls in two places.

tilleryd: Very well, I defer to you guys on this one.

PR #392, opened 2019-11-14
Optional Voyager ingress

### LINKED ISSUE #387: Update doc for non-TCP passthrough ingress controller
Our docs assume we use Voyager for ingress into the mesh and leave it at that. However, if a customer chooses to use something like Nginx, a non-TCP passthrough that supports SNI, then there a few steps we should document.

1. Create a TLS secret https://kubernetes.io/docs/concepts/services-networking/ingress/#tls
2. Apply Nginx YAML using TLS secret https://www.nginx.com/products/nginx/kubernetes-ingress-controller

### LINKED ISSUE #388: Voyager should be optional for k8s deployments
Our edge template assumes using `voyager` ingress controller. For deployments that use a different ingress controller, this needs to be overridden.

```
  annotations:
    kubernetes.io/ingress.class: "voyager"
```

The above annotation should be wrapped with a condition using a value from `greymatter.yaml` to take an ingress class from this list of controllers https://kubernetes.io/docs/concepts/services-networking/ingress-controllers/.

tilleryd:

Resolves #387 
Resolves #388 

Voyager ingress controller doesn't need to be set up if a customer is not using it. We were defaulting to it for all k8s installs. We recently had a customer that wanted to use ingress-nginx so installing Voyager had no value, and if the `appscode` repo isn't installed first, then the helm install would throw an error. This PR makes Voyager ingress optional for k8s.

Added a new global property `k8s_use_voyager_ingress`.

Yes

    1. Update requirements.yaml to use the local `edge` chart.
    2. `helm dep up greymatter`
    3. Without Voyager (you won't have an exposed edge...)
    ```
    helm dep up greymatter
    minikube start -p gm-deploy --memory 6144 --cpus 6
    helm install greymatter -f greymatter.yaml -f greymatter-secrets.yaml -f credentials.yaml --set 
    global.environment=kubernetes -n gm-deploy
    ```
    4. With Voyager
    ```
    helm del --purge gm-deploy
    make minikube
    ```


docs/Ingress.md
```diff
@@ -46,4 +40,12 @@ To verify that Voyager has started, run:
...
+
+### OpenShift
+
+In OpenShift, ingress is defined as a Route with the URL described above. No additional steps are required as the OpenShift router will automatically handle all traffic for you.
+
+## Other ingress
+
+Another popular ingress controller is [ingress-nginx](https://kubernetes.github.io/ingress-nginx/deploy/). If you configure it as a passthrough it'll still handle TLS termination. In which case you will need to use the same certificates from the Grey Matter edge service. One thing to keep in mind, a proper Kubernetes [TLS secret](https://kubernetes.io/docs/concepts/services-networking/ingress/#tls) will need to be created and leveraged by `ingress-nginx`.
```
tilleryd: @chrisbsmith I feel like I'm really botching this explanation. Recommendations are welcome.

docs/Ingress.md
```diff
@@ -1,32 +1,26 @@
...
-This was giving us some pretty cryptic errors, so we decided to just use a L4 TCP proxy that would proxy the entire connection and wouldn't care about TLS. We decided on Voyager, a Kubernetes Ingress controller built on top of HAProxy, mainly for its ease of use and existing integrations with major cloud providers. These integrations are what allow the automatic provisioning of real cloud load balancers from the Kubernetes created `LoadBalancer` resources.
+### Kubernetes
 
-At this writing there is [an issue](https://github.com/appscode/voyager/issues/1415) specifying voyager ingress as a dependency, so we need to manually configure Voyager ingress locally before launching our Grey Matter cluster. This can be done with following commands:
+As of this writing, there is [an issue](https://github.com/appscode/voyager/issues/1415) specifying Voyager ingress as a dependency, so we need to manually configure it locally before launching our Grey Matter cluster. In the below commands, set PROVIDER to the appropriate [cluster provider](https://appscode.com/products/voyager/v11.0.1/setup/install/) for your environment before running. Then run the commands:
```
chrisbsmith:
```suggestion
As of this writing, there is [an issue](https://github.com/appscode/voyager/issues/1415) specifying Voyager ingress as a dependency, so we need to manually configure it locally before launching our Grey Matter cluster. In the below commands, set `PROVIDER` to the appropriate [cluster provider](https://appscode.com/products/voyager/v11.0.1/setup/install/) for your environment before running. Then run the commands:
```

docs/Ingress.md
```diff
@@ -46,4 +40,12 @@ To verify that Voyager has started, run:
   kubectl --namespace=kube-system get deployments -l "release=voyager-operator, app=voyager"
 ```
 
-Now you're all set. When you deploy the edge service, voyager-operator will create a custom `Ingress` resource which will provision a load balancer for you. You can run `kubectl get svc voyager-edge` to see the cluster ip and port.
+Now you're all set. When you deploy the edge service, voyager-operator will create a custom `Ingress` resource. For Kubernetes, a [NodePort](https://kubernetes.io/docs/concepts/services-networking/service/#nodeport) service type will be deployed, which exposes the service on the edge node's IP and static port. For, the EKS cloud provider, a [LoadBalancer](https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer) service type will expose the service using EKS's load balancer. You can run `kubectl get svc voyager-edge` to see the cluster IP and port.
```
chrisbsmith:
```suggestion
Now you're all set. When you deploy the Edge service, the voyager-operator will create a custom `Ingress` resource. For Kubernetes, a [NodePort](https://kubernetes.io/docs/concepts/services-networking/service/#nodeport) service type will be deployed, which exposes the service on the edge node's IP and static port. For, the EKS cloud provider, a [LoadBalancer](https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer) service type will expose the service using EKS's load balancer. You can run `kubectl get svc voyager-edge` to see the cluster IP and port.
```

docs/Ingress.md
chrisbsmith: @tilleryd A few thoughts here: 
* I thought we configured the nginx ingress controller for pure pass-through and not termination.  I thought there was an issue forwarding user credentials along to the Edge
* In order to do ^, we should probably also document that the ingress-nginx controller needs to be configured with the `--enable-ssl-passthrough` commandline option; or the configMap needs to be updated
* I think we should list the annotations we determined were required for nginx ingress to work for Grey Matter.  I don't remember all of them, but I know this one was required.  Maybe we can reach out to Dan and Blaine to see if they can share the annotations with us.
  ```yaml
  nginx.ingress.kubernetes.io/ssl-passthrough: true
  ```
* This sentence is a bit confusing `In which case you will need to use the same certificates from the Grey Matter edge service`.  I think you're trying to say that if the ingress-nginx controller is set to terminate TLS, then it should use the Edge certificates for termination, and then use the same Edge certs to re-encrypt and send the request to the Edge.

Good reminder on requiring the TLS secrets 👍

greymatter.yaml
```diff
@@ -1,5 +1,7 @@
 global:
   environment: openshift
+  # If `environment: kubernetes`, set to true to install Voyager ingress 
+  k8s_use_voyager_ingress: false
```
chrisbsmith: I have no problem with this variable, but it's directly opposed to the camel case issue we discussed a while back, https://github.com/DecipherNow/helm-charts/issues/236.  The bigger thing is that we should just pick a style and stick to it.

greymatter/values.yaml
```diff
@@ -1,5 +1,7 @@
 global:
   environment: openshift
+  # If `environment: kubernetes`, set to true to install Voyager ingress
```
chrisbsmith: Should we add a little more definition to this comment?

```suggestion
  # If `environment: kubernetes`, set to true to install Voyager ingress.  If left `false`, then no ingress will be automatically installed by Helm and you will need to manually add an ingress configuration to your Kubernetes ingress controller.
```

greymatter.yaml
tilleryd: I am sticking with the style of variables around it. We should do a sweeping update en mass.

chrisbsmith: 👍

docs/Ingress.md
tilleryd: @chrisbsmith I pushed an update with the YAML provided by Dan. I stripped out all the commented out lines except for the annotation for passing certs to the upstream edge sidecar. I now can no longer recall why we had to create a k8s TLS secret because it's not being used by this config.

chrisbsmith: I believe we made the TLS secrets with the idea that nginx would do termination and then re-encrypt  to edge.  Since we have the configuration set to pass-through, it is not needed.

PR #393, opened 2019-11-15
Sts data and control api

whenhellfreezes:

Changing data and control api to be statefulsets instead of deployments so that you can do live updates easier. (For data when s3=false)

<br/><br/>

None.

<br/><br/>

NA

<br/><br/>

Ran it in minikube successfully.

<br/><br/>


chrisbsmith:
@whenhellfreezes do you have any idea how an "upgrade" for this PR would work?  Say I already have GM installed with a prior chart, what happens to the deployments for data and gm-control-api?

moosecanswim:
> @whenhellfreezes do you have any idea how an "upgrade" for this PR would work? Say I already have GM installed with a prior chart, what happens to the deployments for data and gm-control-api?

The pvc will remain after the delete.  We may need to change the chart up for this case though.  aka, if there is a pvc that we want to use specify the name, otherwise create a new one.
(https://stackoverflow.com/questions/53186601/re-use-persistentvolume-after-re-installing-a-helm-deployment)

whenhellfreezes:
I will test a helm upgrade with and without chart version increments to see exactly how this behaves.

whenhellfreezes:
It will require a helm del and helm install to transition.

whenhellfreezes:
Merged in master to make sure everything checks out and upped the version numbers. Note that I set gm-control-api from 2.2.1 to 2.1.3. The 2.2.1 was likely a typo.

PR #399, opened 2019-11-20
attempt to build with circleci

chrisbsmith:
Unfortunately #390 is useless because we cannot run GitHub Actions in our organization.  This PR introduces a CircleCI build config that brings our CI back to par with what was happening in Jenkins.  

The CircleCI job is set to only build on PRs (though I'm not sure if that means it will be when a PR is submitted and then again when a PR is merged - this setting may need to be adjusted).

@moosecanswim @tilleryd @kaitmore would definitely like a review since I think you guys know the Helm build process fairly well.


chrisbsmith:
As an FYI - circle did kick off when I submitted the PR.

chrisbsmith:
@tilleryd approving so we can try to get the backlog moving

PR #411, opened 2019-11-25
add gm-data securityContext supported by EKS

tilleryd:

Added gm-data `securityContext` to support non-Openshift deployments. EKS for example, has stricter security policies that require the pod to use the same user/group as the container.

In greymatter.yaml, data securityContext is enabled by default, but I didn't add the user/group values, which can be found in data/templates/data.yaml.

n/a

Deploy to EKS as well as Minikube, it should work on both. If you disable the securityContext then  the data container will not start in EKS.

### DIFF
--- a/data/templates/data.yaml
+++ b/data/templates/data.yaml
@@ -38,6 +38,12 @@ spec:
 {{ toYaml .Values.data.resources | indent 10 }}
         {{- end }}
         {{- if and .Values.global.environment (ne .Values.global.environment "openshift") }}
+        {{- if .Values.data.securityContext.enabled }}
+        securityContext:
+          runAsUser: {{ .Values.data.securityContext.runAsUser }}
+          runAsGroup: {{ .Values.data.securityContext.runAsGroup }}
+          fsGroup: {{ .Values.data.securityContext.fsGroup }}
+        {{- end }}
         securityContext:
           runAsUser: 666
           runAsGroup: 666
--- a/data/templates/data.yaml
+++ b/data/templates/data.yaml
@@ -15,6 +15,12 @@ spec:
       labels:
         app: {{ .Values.data.name }}
     spec:
+      {{- if and .Values.global.environment (ne .Values.global.environment "openshift") }}
+      securityContext:
+        runAsUser: 2000
+        runAsGroup: 1000
+        fsGroup: 2000
+      {{- end }}
       serviceAccountName: {{ .Values.global.waiter.serviceAccount.name }}
       initContainers:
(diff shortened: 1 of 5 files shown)


data/templates/data.yaml
```diff
@@ -15,6 +15,12 @@ spec:
       labels:
         app: {{ .Values.data.name }}
     spec:
+      {{- if and .Values.global.environment (ne .Values.global.environment "openshift") }}
+      securityContext:
+        runAsUser: 2000
+        runAsGroup: 1000
```
chrisbsmith: Our testing found that we needed to use `666` since that's what Data sets.  

https://github.com/DecipherNow/gm-data/blob/2ea984e2c97be6860aed03735e4e1000fc56e1c6/Dockerfile#L60-L61

tilleryd: I linked the dependent PR in the description, and set this to WIP until that's merged in and released.

chrisbsmith: Totally missed that.  Sorry.

chrisbsmith:
Actually @tilleryd, rather than predefining the `securityContext` in the template, I think we should follow the process that other charts use - allow the user to define the context in the values.yaml file and then inject it into the yaml.  This seems cleaner and allows us to change the default values more easily (ie - when [#541](https://github.com/DecipherNow/gm-data/pull/541) is closed

https://github.com/helm/charts/blob/906620c052209419ddc33211b899efcf2088f118/stable/mariadb/templates/slave-statefulset.yaml#L43-L47

https://github.com/helm/charts/blob/906620c052/stable/mariadb/values.yaml#L97-L103

tilleryd:
@chrisbsmith Good call, I'll make that change.

data/templates/data.yaml
tilleryd: I've opted to keep the user and group the same for now so this is OBE.

tilleryd:
@chrisbsmith securityContext can now be set from values

data/values.yaml
```diff
@@ -35,6 +35,11 @@ data:
   pvc:
     #size:
     mountPath: /app/buckets/{{ $.Values.data.aws.bucket}}/{{ $.Values.data.envvars.aws_s3_partition.value }}
+  securityContext:
```
chrisbsmith: It’s becoming painfully obvious that I didn’t document the variables when I initially wrote these charts. Rather than waiting until a mass update, I think it best if e we start including comments to any new variables we add to the values file. Even though the securityContext is obvious, can you please add a comment to the new values?

tilleryd: Done

PR #415, opened 2019-11-29
Set catalog config polling interval

### LINKED ISSUE #414: Set catalog config_polling_interval
To mitigate the issue [described here](https://github.com/DecipherNow/gm-catalog/issues/248), we should add the `config_polling_interval` variable to catalog to avoid entries being overwritten by Data. I propose setting it to 30s to give the catalog init job a good amount of time to complete.

**This is not a long term solution** - if for some reason a POST happens at the exact moment catalog reads from data, the cluster will be overwritten. Adding `config_polling_interval` should make this less likely though.

bcmendoza:

Sets the `config_polling_interval` setting in Catalog to `30s` to prevent race conditions when reading from Data and writing to Catalog's internal model.
Closes #414 

None.

None needed.

In the `requirements.yaml` file, the following must be changed to use the local file for Catalog's repository:
```
  - name: catalog
    version: '2.1.5'
    repository: 'file://../catalog'
```


tilleryd:
@bcmendoza Is this an interim or long term solution?

bcmendoza:
@tilleryd This is an interim solution. We need to discuss how to manage Catalog's reads/writes to Data. Writes should block reads.

PR #428, opened 2019-12-04
Change how Edge Ingress is Defined

chrisbsmith:
# Edge Ingress Updates

This PR makes very significant updates to how Edge Ingress can be defined.  Currently, we only offer two forms of ingress: OpenShift Routes or Voyager support with Kubernetes.  There are many other ingress controllers for Kubernetes and Voyager may not always be the flavor of choice for a customer.  Additionally, when deployed to EKS, the Voyager ingress creates a new ELB everytime the edge pod is cycles, essentially making it impossible to properly DNS configure.

## PR Details

This PR changes how we define Kubernetes ingress.  Now, we give the users the option to provide their ingress annotations.  Users can define these new variables

`.edge.ingress.apiVersion`: This is the Kubernetes API version to use in the resource definition
`.edge.ingress.annotations`: These are any specific annotations the ingress resource will need.  For example, both Voyager and NGINX Ingress use specific annotations to configure the ingress controller
`edge.ingress.rules`: These are the rules the ingress controller will use to route traffic to the Edge pod

Voyager has been left as the default ingress controller but an example for NGINX ingress has been provided in the Ingress.md

## greymatter.yaml updates

The updates to the greymatter.yaml are address above and can be found in the `edge/values.yaml` file

## Testing

This has been tested in the new Development environment I'm working on, but it should be thoroughly tested for Voyager and Minishift.

@tilleryd @kaitmore @dgoldstein1 @whenhellfreezes


whenhellfreezes:
Ran successfully in minikube.

edge/values.yaml
```diff
@@ -41,9 +47,52 @@ edge:
...
+    
+    # Provide the rule set you want your ingress controller to use
+    rules:
+      - tcp:
+        port: '80'
+        backend:
+          serviceName: edge
+          servicePort: 8080
+      - tcp:
+        port: '443'
```
kaitmore: Could we add back `nodePort: "30000"` for this block and `nodePort: "30001"` for the one above here? Training materials rely on the ports being pinned.

chrisbsmith: Thanks for catching this.  I've added them back in.

edge/values.yaml
```diff
@@ -41,9 +47,54 @@ edge:
...
+
+    # rules:
+    #   - host: greymatter.development.deciphernow.com
+    #     http:
+    #       paths: 
+    #         - path: /
+    #           backend:
+    #             serviceName: edge
+    #             servicePort: 8080
+
```
dgoldstein1: do we need this here if it's also in the `docs` folder?

chrisbsmith:
>Can you make a matrix in the docs/ingress of the types we currently support?

I don't think this is necessary.  The ingress doc already states that we support two types of ingress: Openshift Routes and Kubernetes Ingress.

>Does this PR fix this?? We need to test that these changes work on a new eks cluster

This PR does not attempt to fix this voyager issue, nor do I think it should.  When running in EKS, our recommendation will be to use an NLB ingress controller and an nginx ingress controller.

>Unless I'm mistaken, I don't think we've ever got minishift running.. did you mean minikube?

You are correct, I meant minikube.

edge/values.yaml
chrisbsmith: It can.  I left it there as an example for anyone that looks at the values.yaml file first.  Open to removing it if we don't want it here.

dgoldstein1: I don't have a strong preference one way or the other.. want to make sure you didn't forget to delete

chrisbsmith:
@dgoldstein1  actually, I need to increment the version of the Grey Matter chart so we don't over write 2.1.3.

@tilleryd @kaitmore would love another set of eyes on this if you can

PR #430, opened 2019-12-05
fix data securityContext

tilleryd:

Remove bad conditional logic in data.yaml securityContext.

No

n/a

1. Ran `make minikube` and confirm the security context by running `kubectl get -l app=internal-data -o yaml`

### DIFF
--- a/data/templates/data.yaml
+++ b/data/templates/data.yaml
@@ -44,10 +44,6 @@ spec:
           runAsGroup: {{ .Values.data.securityContext.runAsGroup }}
           fsGroup: {{ .Values.data.securityContext.fsGroup }}
         {{- end }}
-        securityContext:
-          runAsUser: 666
-          runAsGroup: 666
-          fsGroup: 666
         {{- end }}
         env:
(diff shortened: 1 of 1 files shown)


data/templates/data.yaml
```diff
@@ -15,6 +15,8 @@ spec:
       labels:
         app: {{ .Values.data.name }}
     spec:
+      securityContext:
+        fsGroup: 666
```
chrisbsmith: You should actually be using the same template as before `{{ .Values.data.securityContext.fsGroup }}` rather than hardcoding the value here.

data/templates/data.yaml
```diff
@@ -44,10 +46,6 @@ spec:
           runAsGroup: {{ .Values.data.securityContext.runAsGroup }}
           fsGroup: {{ .Values.data.securityContext.fsGroup }}
```
chrisbsmith: Since `fsGroup` isn't valid in the container spec, we should remove this line.

tilleryd: Man, my attention to detail was lacking in this PR...

Issue #408, opened 2019-11-22
Roll back catalog


kaitmore:
We need to roll back catalog to version 1.0.1 to resolve various bugs, see 
https://github.com/DecipherNow/helm-charts/issues/376

whenhellfreezes:
Issue kept so that we can remember to cut a new 1.0.3 version.

PR #431, opened 2019-12-06
Catalog version bump 1.0.3

dgoldstein1:

see https://github.com/DecipherNow/gm-catalog/issues/250

- update catalog to v1.0.3
- update all proxies to version 1.1.1

- setting `zone` to be the same value for all proxies and gm-control-api, due to https://github.com/DecipherNow/gm-catalog/issues/250

- not yet, but may not need to

- dry run testing, making sure that configs are valid

```bash
david@david-Oryx-Pro:~/dev/work/helm-charts$ mv greymatter/requirements-dev.yaml greymatter/requirements.yaml
david@david-Oryx-Pro:~/dev/work/helm-charts$ helm dep up greymatter
Hang tight while we grab the latest from your chart repositories...
...Unable to get an update from the "local" chart repository (http://127.0.0.1:8879/charts):
	Get http://127.0.0.1:8879/charts/index.yaml: dial tcp 127.0.0.1:8879: connect: connection refused
...Successfully got an update from the "appscode" chart repository
...Successfully got an update from the "decipher" chart repository
...Successfully got an update from the "stable" chart repository
Update Complete.
Saving 11 charts
Deleting outdated charts
david@david-Oryx-Pro:~/dev/work/helm-charts$ helm install greymatter -f greymatter.yaml -f greymatter-secrets.yaml --name gm --dry-run
NAME:   gm
```
- made this change
```
--- a/greymatter/requirements.yaml
+++ b/greymatter/requirements.yaml
@@ -1,7 +1,7 @@
 dependencies:
   - name: catalog
     version: '2.1.6'
-    repository: 'https://nexus.production.deciphernow.com/repository/helm-hosted'
+    repository: 'file://../catalog'
 
   - name: data
     version: '2.1.4'
```
and then booted up minikube successfully 
- TODO: make sure changes to mesh using greymatter cli are applied successfully

_marking this WIP because the `release/2.1` branch is currently not stable, so cannot test this against it until those errors are fixed_


bcmendoza:
In `greymatter/values.yaml` line 65, I'm reading `version: 1.0.0` for `sidecar`. This should be updated to `1.1.1`.

greymatter.yaml
```diff
@@ -93,6 +93,9 @@ global:
       proxy_dynamic:
         type: 'value'
         value: 'true'
+      xds_zone:
+        type: 'value'
+        value: '{{ .Values.global.zone }}'
```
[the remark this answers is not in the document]
chrisbsmith: Before we make this change, we need to make sure that we understand how the templating it working.  I agree that everything needs to be the same (`zone-default-zone`), but when deploying for a customer with our defaults, I get the following output.

image

kaitmore:
@bcmendoza @dgoldstein1 

I'm seeing a ton of errors from gm-control-api-init about invalid keys:

```
[info] 2019/12/09 16:53:28 Preferring --api.key for authentication
create: {
  "message": "Route().Create failed: validateKeys: invalid DomainKey: domain-data-internal",
  "code": "MiscCode"
}

[info] 2019/12/09 16:53:28 Preferring --api.key for authentication
create: {
  "message": "Route().Create failed: validateKeys: invalid DomainKey: domain-data-internal",
  "code": "MiscCode"
}
```

greymatter.yaml
```diff
@@ -699,17 +702,17 @@ internal-jwt:
...
-  image: 'docker.production.deciphernow.com/deciphernow/gm-proxy:1.0.0'
+  image: 'docker.production.deciphernow.com/deciphernow/gm-proxy:1.1.1'
   create_sidecar_secret: true
 
 edge:
   sidecar:
-    version: 1.0.0
-    image: 'docker.production.deciphernow.com/deciphernow/gm-proxy:1.0.0'
+    version: 1.1.1
+    image: 'docker.production.deciphernow.com/deciphernow/gm-proxy:1.1.1'
```
tilleryd: What does updating all sidecars to 1.1.1 get us? If we upgrade we have to test all the changes since 1.0.0.

dgoldstein1: see the info for this pr ^^

https://github.com/DecipherNow/gm-catalog/issues/250

chrisbsmith:
@tilleryd would you agree that changing the proxy from `1.0.0` to `1.1.1` would be sufficient to cut a release?  If so, it would be a good time to start using the `-dev` tag.

dgoldstein1:
created related issue https://github.com/DecipherNow/helm-charts/issues/441

kaitmore:
@bcmendoza you need to remove quotes around the templates, e.g:

'{{ .Values.global.zone }}' should just be {{ .Values.global.zone }}

dgoldstein1:
@kaitmore I made this change. Where are you referring?

As far as I can tell, all the current variable insertion have a single quote on both sides, i.e.

```yaml
    gm_control_api_persister_path:
      type: value
      value: '{{ .Values.gmControlApi.pvcMountPoint }}'
...
```

kaitmore:
@dgoldstein1 sorry about that! should have looked closer before I commented. You're right there should be quotes for values files. I thought it was a template file which doesn't use quotes.

dgoldstein1:
@kaitmore ok, no worries! It was very possible I made a mistake with this ^^

kaitmore:
Resolves #408

dashboard/values.yaml
```diff
@@ -128,6 +128,9 @@ sidecar_prometheus:
     xds_cluster:
       type: 'value'
       value: 'prometheus'
+    xds_zone:
```
chrisbsmith: Rather than declaring the `xds_zone` variable in every sidecar for every service, you should put it in the `global.sidecar.envvars` section of `greymatter.xml` so that the `greymatter.envvars` template will pick them up. Variables declared in the individual files like this would only be overrides to the defaults.

Template:  https://github.com/DecipherNow/helm-charts/blob/release-2.1/dashboard/templates/_envvars.tpl#L22-L45 

Sidecar envvars: https://github.com/DecipherNow/helm-charts/blob/release-2.1/greymatter.yaml#L85-L111

This should be changed in all files throughout this PR.

catalog/values.yaml
```diff
@@ -72,7 +72,7 @@ catalog:
       value: '{{ .Values.global.control.name }}.{{ .Release.Namespace }}.svc.cluster.local:{{ .Values.global.control.port }}'
     control_server_0_zone_name:
       type: value
-      value: '{{ .Values.global.zone }}'
+      value: 'zone-default-zone'
```
chrisbsmith: Why did you choose to hard code this value rather than use the value defined at `global.zone`?

control/values.yaml
```diff
@@ -51,7 +51,7 @@ control:
       value: 'xxx'
     gm_control_api_zone_name:
       type: 'value'
-      value: 'default-zone'
+      value: 'zone-default-zone'
```
chrisbsmith: Is this another place where `global.zone` should be used?

data/templates/mongo.yaml
```diff
@@ -17,7 +17,7 @@ spec:
       labels:
         app: {{ .Values.mongo.name }}
     spec:
-      {{- if and .Values.global.environment (eq .Values.global.environment "eks") }}
+      {{- if and .Values.global.environment (eq .Values.global.provider "eks") }}
```
chrisbsmith: `global.provider` appears to be new with this PR, but I don't see where it's been defined or discussed anywhere.

catalog/values.yaml
```diff
@@ -63,7 +63,7 @@ catalog:
       value: 'gmdata'
     config_polling_interval:
       type: value
-      value: '30s'
+      value: '300s'
```
tilleryd: @dgoldstein1 Have you tested with 60s, 120s? 300s (5 minutes) is a long time to wait for an update.

dashboard/values.yaml
tilleryd: Also `value: 'zone-default-zone'` should be replaced with `value: $.Values.global.zone`.

catalog/values.yaml
dgoldstein1: Yes, I tested with different values-- the race condition is alleviated at 300 seconds. We don't "wait" for an update on every config poll interval. This is just the interval where catalog syncs with data. When we set this at something lower (i.e. 10s), we get more frequent overwrites and lost data. 

https://github.com/DecipherNow/greymatter/issues/236

dgoldstein1:
@chrisbsmith @tilleryd 

The hard-coded `zone` has been an issue for a while.. some parts have always been hard coded and some have used `global.zone`. It's outside the scope of this PR to fix this here since there's already a bunch going on. For now Brian and I agreed it would be easier to hard code everything to `zone-default-zone` and then fix this in a later pull request.

https://github.com/DecipherNow/helm-charts/issues/441

data/templates/mongo.yaml
dgoldstein1: I needed to add an additional parameter here to deploy to eks.. but I'll fix this in a separate PR instead of clouding this one

https://github.com/DecipherNow/helm-charts/issues/449#issuecomment-566716438

dgoldstein1:
@chrisbsmith @tilleryd 

Thanks for the feedback. Hard-coding zone has been an issue for a while. Changing the zone to anything except `zone-default-zone` has never worked. It's a heavy lift to fully replace. Brian and I decided to create https://github.com/DecipherNow/helm-charts/issues/441 as a separate issue since our changes were getting outside the scope of this PR and we thought that kind of change should be done in one fell swoop.. 

I've gone ahead an made a few changes, including deleting `zone` from greymatter.yaml so that now we know all zones hard-coded right now.

catalog/values.yaml
tilleryd: Did you test 3 mins and 4 minutes? How low did you go before noticing issues?

dgoldstein1: The issue we're experiencing only occurs when we send an API request to catalog when it's in the middle of writing or reading from GM Data. Bumping the polling interval to 3,4, or 5 minutes simply lessens the chance that this will occur. We could still "notice" issues at 5 minutes, although it's less likely.

https://github.com/DecipherNow/gm-catalog/issues/261

data/templates/mongo.yaml
chrisbsmith: I'm good with adding this one here if it's necessary; just document it.

chrisbsmith:
> Changing the zone to anything except `zone-default-zone` has never worked. It's a heavy lift to fully replace.

Can you explain why this is a heavy lift?  I'm not sure why you couldn't just use `.global.zone` where you've used `zone-default-zone`.  This would eliminate the need for a future PR.

Also, please review my [comment](https://github.com/DecipherNow/helm-charts/pull/431#discussion_r362509329) above.  The Helm Charts use a template for the sidecar environment variables and that's where the new `xds_zone` variable should be placed. This will ensure that any sidecar deployed by the helm charts will have the proper variables.  Any variable defined in the individual chart values files would then be an override to the defaults.  This will also reduce the number of files you're changing since you won't have to add the `xds_zone` to each chart.

dgoldstein1:
@chrisbsmith 

- Updating the zone for the whole helm-charts repo requires a bunch of changes since it's used in a bunch of different types of places: every sidecar, every configuration object in the bootstrap, in some data script configs. This would definitely be doable, but as I said ^^ Brian and I ran into problems with this (you can look at the commits and see that we actually did this and then reverted) and it will be easier to debug and have better git etiquette if we make another PR for this as a separate issue. 
- I can definitely put zone into the global sidecar vars-- I just missed that

chrisbsmith:
@dgoldstein1 
Thanks for the description.  Go ahead and make the updates for the sidecar envvars and plan to update back to `global.zone` in the next PR.

dgoldstein1:
@chrisbsmith I've made sure that `xds_zone` is coming from the global greymatter sidecar values for each chart.

chrisbsmith:
Thanks @kaitmore

PR #457, opened 2020-01-02
WIP Optionally install oAuth configuration

tilleryd:

This PR allows users to install Grey Matter with oAuth configured. Two ingresses will be setup, one for oAuth and one for PKI. In the oAuth scenario, the only service that is configured in the mesh for the standard PKI edge will be gm-control-api. This allows users to continue to hit the API via the CLI from the edge.

In greymatter.yaml, there are now two new edge ingress configurations using Voyager and an additional boolean to enable oAuth in the mesh.

In greymatter-secrets.yaml, there are new oAuth attributes which will be used to configure the edge-oauth proxy in the mesh.

Will be in a follow-up PR.


edge/templates/edge-ingress.yaml
```diff
@@ -29,5 +29,7 @@ metadata:
 spec:
   rules:
 {{ toYaml .rules | indent 4 }}
+  tls:
```
chrisbsmith: This assumes that TLS will always be provided.  In an nginx-ingress configuration, we enable nginx to do tls passthrough, so this the `tls` block is not needed.

Recommend adding a check here to see if the `tls` config option exists in the values file first.

edge/templates/edge-oauth-ingress.yaml
```diff
@@ -0,0 +1,38 @@
...
+kind: Ingress
+metadata:
+  name: edge-oauth
+  namespace: {{ $.Release.Namespace }}
+  annotations:
+{{ toYaml .annotations | indent 4 }}
+spec:
+  rules:
+{{ toYaml .rules | indent 4 }}
+  tls:
```
chrisbsmith: My tls comment from above would hold true here too.

edge/templates/edge.yaml
```diff
@@ -17,14 +17,14 @@ spec:
     spec:
       containers:
       - name: edge
-        image: {{ tpl .Values.sidecar.image $ | quote }}
-        imagePullPolicy: {{ .Values.sidecar.imagePullPolicy }}
+        image: {{ tpl .Values.edge.sidecar.image $ | quote }}
```
chrisbsmith: This is interesting because we chose to use `sidecar.image` since edge is nothing but a sidecar configured as an edge.  However, I see that we also wrote `edge.sidecar.image`.  Part of me wants to get rid of `edge.sidecar.image` as it's just another place we need to keep track of a sidecar version.  I'd expect the edge and sidecar versions to always be in sync.  Thoughts?

gm-control-api/config/bootstrap.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env sh
 
-MESH_CONFIG_DIR="/etc/config/mesh/"
+MESH_CONFIG_DIR="/etc/config/mesh"
```
chrisbsmith: Have you tested without the trailing `/`?

gm-control-api/json/special/edge-oauth/proxy.json
```diff
@@ -0,0 +1,28 @@
...
+  "proxy_key": "edge-oauth-proxy",
+  "domain_keys": ["edge-oauth"],
+  "listener_keys": ["edge-oauth-listener"],
+  "name": "edge-oauth",
+  "listeners": null,
+  "active_proxy_filters": ["gm.metrics", "gm.oauth"],
+  "proxy_filters": {
+    "gm_oauth": {
+      "provider": "https://deciphernow-davidtillery.okta.com",
+      "client_id": "0oa28nhtmfWgx0hTu357",
```
chrisbsmith: These probably should be secrets or env vars that can be substituted into place.

gm-control-api/templates/configmaps.yaml
```diff
@@ -1,18 +1,18 @@
...
 {{- $svidList = append $svidList (print "spiffe://" $top.Values.global.spire.trustDomain "/spire-client/mTLS") }}
 {{- /* We loop over all the services to create the individual service objects */}}
 {{- range $name, $service := .Values.global.services }}
 {{- if $service.controlEnabled }}
----
 {{/* {{- $lSecret := (default $service.removeListenerSecret false) }} */}}
 {{- $serviceExtras := (dict "top" $top "authorizedSvids" $svidList "mTLS" $top.Values.global.spire.enabled "zone" $top.Values.gmControlApi.zone "svidList" (toJson $svidList)) }}
 {{- $svc := merge $service $serviceExtras }}
 {{- $scope := set $top "service" $svc }}
+---
```
chrisbsmith: Why did you move this break?  I'm not sure of the impact it may have.

greymatter.yaml
```diff
@@ -701,60 +704,104 @@ sidecar:
...
-      # For instance, these values will work for nginx ingress
+      #################
+      # NGINX INGRESS #
+      #################
+      apiVersion: extensions/v1beta1
+      annotations:
+        kubernetes.io/ingress.class: "nginx"
+        nginx.ingress.kubernetes.io/ssl-passthrough: "true"
+      rules:
+        - host: pki.dtillery-ingress.com
```
chrisbsmith: don't forget that you have "personal" info in this file.

edge/templates/edge-ingress.yaml
tilleryd: removed in favor of using Voyager with SSL passthrough

edge/templates/edge-oauth-ingress.yaml
tilleryd: removed in favor of using Voyager with SSL passthrough

edge/templates/edge.yaml
tilleryd: Agreed, but that's out of scope of this PR.

gm-control-api/config/bootstrap.sh
tilleryd: Yes, all subsequent string concatenations led to double forward slashes before this change.

gm-control-api/templates/configmaps.yaml
tilleryd: It's just for readability. The break is now directly above the following YAML.

tilleryd: I tested everything and it works fine.

greymatter.yaml
tilleryd: OBE

chrisbsmith:
@tilleryd This has been a WIP for quite a long time now.  At this point, I don't think that we should push it forward to the `release-2.1` branch. What do you think?

tilleryd:
This is OBE due to the new filters and helm-charts restructure.

PR #458, opened 2020-01-02
Update service account documentation

### LINKED ISSUE #456: Update service account documentation
Current service account documentation (https://github.com/DecipherNow/helm-charts/blob/release-2.1/docs/Service%20Accounts.md) is sufficient for deploying Grey Matter, but does not provide any insight for service accounts when deploying services to a different namespace.

Specifically, need to document that additional service accounts don't need to be created, rather only new `RoleBindings` need to be created to bind the service account to the new namespace

chrisbsmith:
This PR updates the Service Account documentation to add details on how to enable Grey Matter the ability to monitor services in other, non default, namespaces.

closes #456

### DIFF
--- a/docs/Service Accounts.md
+++ b/docs/Service Accounts.md
@@ -10,6 +10,8 @@ Various services in the mesh require Kubernetes Service Accounts to function. Th
   - `gm-control` - waits for `gm-control-api` to be up - see above, also needs access to pods, so make sure you create one service account for this.
   - `gm-control-api-init` - waits for `gm-control-api` to be up to bootstrap the mesh configuration
 
+## Using Helm to Create the Service Accounts
+
 The waiter service account can be created automatically with `.Values.global.waiter.serviceAccount.create` set to `true`. Otherwise, all services that need access to a waiter service account will use the one specified by `.Values.global.waiter.serviceAccount.name`
 
 All of the service accounts needed for Grey Matter can either be created automatically by Helm (if it has the appropriate permissions), or be created manually by a cluster admin. This is configured in the `serviceAccount` map that is found at different locations for various service accounts, which always looks like this:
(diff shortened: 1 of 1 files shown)


docs/Service Accounts.md
```diff
@@ -55,4 +59,55 @@ If you're deploying into an environment where Tiller doesn't have sufficient per
...
+    namespace: greymatter
+roleRef:
+  kind: Role
+  name: waiter-sa-role
+  apiGroup: rbac.authorization.k8s.io
+```
+
+Notice that only the `metadata.namespace` field needs to be updated for both files.  The `waiter-sa` service account has already been created in the `greymatter` namespace and a new `Role` and `RoleBinding` can leverage the existing service account.  
+
+If you manually created the service accounts using the `greymatter-service-accounts.yaml` file referenced above, you will only need to perform this action for the `waiter-sa` account. If you `control-sa` and `prometheus-sa` were created as non-Cluster roles (ie: not `ClusterRole` and `ClusterRoleBinding`), you will need to perform the same steps for those accounts as well.
```
kaitmore:
```suggestion
If you manually created the service accounts using the `greymatter-service-accounts.yaml` file referenced above, you will only need to perform this action for the `waiter-sa` account. If `control-sa` and `prometheus-sa` were created as non-Cluster roles (ie: not `ClusterRole` and `ClusterRoleBinding`), you will need to perform the same steps for those accounts as well.
```

docs/Service Accounts.md
```diff
@@ -55,4 +59,55 @@ If you're deploying into an environment where Tiller doesn't have sufficient per
     oc apply -f greymatter-service-accounts.yaml
     ```
 
+## Creating Service Accounts for Other Namespaces
+
+Grey Matter has the ability to monitor services deployed in any namespace in Kubernetes, but the service accounts need access to the namespace.  In order to enable a service account to have access to another namespace, a new `Role` and `RoleBinding` need to be created.  This is a manual step that mnuch be done for each namespace that you will be deploying Grey Matter monitored services.
```
kaitmore:
```suggestion
Grey Matter has the ability to monitor services deployed in any namespace in Kubernetes, but the service accounts need access to the namespace.  In order to enable a service account to have access to another namespace, a new `Role` and `RoleBinding` need to be created.  This is a manual step that must be done for each namespace that you will be deploying Grey Matter monitored services.
```

chrisbsmith: Thanks for the catches!

PR #463, opened 2020-01-13
Pulldown catalog values

whenhellfreezes:

Helm does actually an okay job at merging a parent charts values with the value.yaml of its children. As such I am pulling values into the children charts to lessen confusion. This PR is for catalog's values.

Where greymatter.yaml differred from catalog/values.yaml I changed catalog/values.yaml. These fields are:
- client_identity
- client_email
- server_0_address
- config_polling_interval

These values are now conformant to the post svc.cluster.local -> svc change and the impersonation changes.

<br/><br/>

This change is backwards compatible with most custom.yaml/greymatter.yaml from the past since those over specify information. The only problem is if greymatter.yaml _didn't_ specify the four fields mentioned. Both greymatter.yaml and the di2e version **did** specify.

We could drop the changes to catalog/values.yaml to make this not an issue.

<br/><br/>

NA

<br/><br/>

Tested in minikube.

<br/><br/>


whenhellfreezes:
Testing was done with `make dev` do I need to update all the version numbers?

catalog/values.yaml
```diff
@@ -63,13 +63,13 @@ catalog:
       value: 'gmdata'
     config_polling_interval:
       type: value
-      value: '300s'
+      value: '30s'
```
kaitmore: FYI I think this was supposed to be 300s, see convo here https://github.com/DecipherNow/helm-charts/pull/431/files#diff-4a4f56282677fc72e58c4d8cb15d1f24R66

whenhellfreezes: I can change it 300s again. Though given that greymatter.yaml was still at 30s it is overriding to a bad value right now on `release-2.1`.

catalog/values.yaml
```diff
@@ -126,10 +126,10 @@ catalog:
       value: 'world'
     client_identity:
       type: value
-      value: 'CN=gm-control,OU=Engineering,O=Decipher Technology Studios,=Alexandria,=Virginia,C=US'
+      value: 'C=US,ST=Virginia,L=Alexandria,O=Decipher Technology Studios,OU=Engineering,CN=*.greymatter.svc.cluster.local'
```
chrisbsmith: This should really be `CN=gm-catalog,OU=Engineering,O=Decipher Technology Studios,=Alexandria,=Virginia,C=US`.  
```suggestion
      value: 'CN=gm-catalog,OU=Engineering,O=Decipher Technology Studios,=Alexandria,=Virginia,C=US'
```

catalog/values.yaml
```diff
@@ -126,10 +126,10 @@ catalog:
...
     client_email:
       type: value
-      value: 'gm-control@deciphernow.com'
+      value: 'engineering@deciphernow.com'
```
chrisbsmith:
```suggestion
      value: 'gm-catalog@deciphernow.com'
```

catalog/values.yaml
chrisbsmith: At least I believe it should, please test with this value.

whenhellfreezes: Will test with your suggestions. What I set it to again is what `release-2.1` is doing right now.

whenhellfreezes: Crashes with your suggestions. :(

whenhellfreezes: Seems like it can connect but then doesn't have permissions with data.

chrisbsmith: Please post the logs.

whenhellfreezes:
Rolled back

chrisbsmith:
@whenhellfreezes it may have failed b/c I improperly told you to use `gm-catalog@deciphernow.com` for the email address.  It looks like we configure the `gm-catalog` persona to use `engineering@deciphernow.com`.  Please update to try using the `gm-catalog` user again but with the engineering email.  

For reference, you can see the users we load into internal JWT below.  The data for the DN has to match the email for JWT to process it properly.

https://github.com/DecipherNow/helm-charts/blob/1a5054887114cdd32e3c7f287e8e0917d6e6f53a/greymatter.yaml#L668-L681

whenhellfreezes:
> @whenhellfreezes it may have failed b/c I improperly told you to use `gm-catalog@deciphernow.com` for the email address. It looks like we configure the `gm-catalog` persona to use `engineering@deciphernow.com`. Please update to try using the `gm-catalog` user again but with the engineering email.
> 
> For reference, you can see the users we load into internal JWT below. The data for the DN has to match the email for JWT to process it properly.
> 
> https://github.com/DecipherNow/helm-charts/blob/1a5054887114cdd32e3c7f287e8e0917d6e6f53a/greymatter.yaml#L668-L681

Worked. Changing only DN and keeping email was successful. Set chart versions as well.

whenhellfreezes:
Took the scenic route to fixing merge conflicts.

whenhellfreezes:
@chrisbsmith can I get another look at this?

PR #464, opened 2020-01-13
Control api sidecar version improvement

whenhellfreezes:

The version of the sidecar for control api should not be determined by catalog.

<br/><br/>

NA

<br/><br/>

NA

<br/><br/>

Tested in Minikube.

<br/><br/>

### DIFF
--- a/gm-control-api/values.yaml
+++ b/gm-control-api/values.yaml
@@ -108,7 +108,7 @@ gmControlApi:
     size:
 
 sidecar:
-  version: '{{- $.Values.global.catalog.sidecar.version | default "1.0.0" }}'
+  version: '{{- $.Values.global.sidecar.version | default "1.1.1" }}'
   image: 'docker.production.deciphernow.com/deciphernow/gm-proxy:{{ tpl $.Values.sidecar.version $ }}'
   imagePullPolicy: IfNotPresent
   resources:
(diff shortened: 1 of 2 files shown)


gm-control-api/values.yaml
```diff
@@ -108,7 +108,7 @@ gmControlApi:
     size:
 
 sidecar:
-  version: '{{- $.Values.global.catalog.sidecar.version | default "1.0.0" }}'
+  version: '{{- $.Values.global.gmControlApi.sidecar.version | default $.Values.global.sidecar.version }}'
```
sofiashalotenko: What does that value actually refer to?  There doesn't seem to be a global.gmControlApi.sidecar.version in the file.

whenhellfreezes: Yeah but the other charts do have those. So this makes it consistent and at worst defaults to something that is definitely there.

sofiashalotenko: I guess my implied suggestion is maybe we add a gmControlApi.sidecar.version option in the yaml so it makes more sense--unless there's some reason not to.

whenhellfreezes: So I added a change. The option is visible in greymatter.yaml but I think it is better to use the global version so I commented it out.

PR #465, opened 2020-01-13
The data_prefix should belong to data

whenhellfreezes:

It makes sense for the prefix of where data lives to be a peice of information owned by data.

<br/><br/>

NA

<br/><br/>

NA

<br/><br/>

Tested in minikube
<br/><br/>


whenhellfreezes:
> @whenhellfreezes The `data_prefix` being used here tells Catalog where it can find Grey Matter data (ie: `127.0.0.1:8080/{data_prefix}`), and is an integral variable for Catalog.
> 
> Rather than making this change, it would probably be efficient to document how this variable is actually being used.

So the prefix is owned by the route which is owned by catalog.

chrisbsmith:
> > @whenhellfreezes The `data_prefix` being used here tells Catalog where it can find Grey Matter data (ie: `127.0.0.1:8080/{data_prefix}`), and is an integral variable for Catalog.
> > Rather than making this change, it would probably be efficient to document how this variable is actually being used.
> 
> So the prefix is owned by the route which is owned by catalog.

That's correct.

PR #479, opened 2020-01-23
remove data sidecar configs

tilleryd:

`data/values.yaml` was updated to include sidecar configs that override global sidecar configs. The xds_host env var was hard coded to `control.default.svc.cluster.local` which meant gm-data wouldn't connect to gm-control in a namespace other than `default`. I removed the configs because this aligns with how we're managing sidecar configs in all the other charts.

none

n/a

1. Deploy to a different namespace other than default
2. Ensure that `catalog -> data -> jwt` works.

### DIFF
--- a/data/Chart.yaml
+++ b/data/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.0.0
 description: A Helm chart to deploy Grey Matter Data
 name: data
-version: 2.1.5
+version: 2.1.6
 icon: https://s3.amazonaws.com/decipher-public/grey-matter/branding/grey-matter-mark-rgb.png
 maintainers:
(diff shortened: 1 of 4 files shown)


whenhellfreezes:
Instead of popping of we should fix in place so less logic has to live in greymatter.yaml

chrisbsmith:
@whenhellfreezes actually, this is the preferred fix for now. We put all of the environment variables that are shared amongst all proxies in the greymatter.yaml. The unique values (proxy name) are kept in the chart/values.yaml file. 

We can investigate moving these configs into greymatter/values.yaml but I don’t think we should in this PR.

tilleryd:
@whenhellfreezes I consciously went for consistency with how the other charts are implemented. This is a fix rather than a refactor.

PR #496, opened 2020-03-04
Liveness probe to ensure XDS HOST is set correctly

whenhellfreezes:

This implements the liveness probe that justincely prototyped out to check that XDS_HOST is correctly set. This probe is on dashboard to make sure that the globals are correct (would be really hard to mess up) and also on data in case it is deployed standalone.

<br/><br/>

None

<br/><br/>

NA
<br/><br/>

Ran in minikube. Checked with using both the correct and wrong namespace for XDS_HOST.
<br/><br/>

### DIFF
--- a/data/templates/data.yaml
+++ b/data/templates/data.yaml
@@ -110,6 +110,14 @@ spec:
               - "curl -k --cacert /etc/proxy/tls/sidecar/ca.crt --key /etc/proxy/tls/sidecar/server.key --cert /etc/proxy/tls/sidecar/server.crt https://localhost:8181/list/1/"
           initialDelaySeconds: 20
           periodSeconds: 5
+        livenessProbe:
+          exec:
+            command:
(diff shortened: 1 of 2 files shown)


data/templates/data.yaml
```diff
@@ -110,6 +110,14 @@ spec:
               - "curl -k --cacert /etc/proxy/tls/sidecar/ca.crt --key /etc/proxy/tls/sidecar/server.key --cert /etc/proxy/tls/sidecar/server.crt https://localhost:8181/list/1/"
           initialDelaySeconds: 20
           periodSeconds: 5
+        livenessProbe:
```
sofiashalotenko: Since this is here in case of data-standalone, maybe this should be conditional based on Values.data.deploy.standalone?

moosecanswim: The liveness probe can be run for either standalone and integrated.

Issue #488, opened 2020-02-21
Setting sidecar.create_sidecar_secret to false did not have intended result


chrisbsmith:
When deploying Grey Matter to a new namespace, the sidecar-certs secret was already created in the namespace, so I set `sidecar.create_sidecar_secret` to `false`. However, when running a `helm install`, the install failed because it still tried to build the sidecar-certs secret.

tilleryd:
Initially reported in https://support.deciphernow.com/a/tickets/180

PR #497, opened 2020-03-05
allows for the situation where the sidecar secret is already installed in the namespace

sofiashalotenko:

This PR is in response to [this issue](https://github.com/DecipherNow/helm-charts/issues/488).  What was happening was that both `create_sidecar_secret` _and_ `mesh_tls.use_provided_certs` had to be false or else the sidecar secret would be created.  I removed the `mesh_tls.use_provided_certs` from the logic--I'm not sure why it was put there in the first place but I think it makes more sense without it.

<br/><br/>

n/a
<br/><br/>

n/a
<br/><br/>

Ran in minikube.  I applied the sidecar secret manually first, then I helm installed greymatter with `create_sidecar_secret` set to false.
<br/><br/>

### DIFF
--- a/greymatter/templates/sidecar-secret.yaml
+++ b/greymatter/templates/sidecar-secret.yaml
@@ -1,4 +1,4 @@
-{{- if or (.Values.sidecar.create_sidecar_secret) (.Values.global.mesh_tls.use_provided_certs) }}
+{{- if .Values.sidecar.create_sidecar_secret }}
 {{- if .Values.sidecar.certificates }}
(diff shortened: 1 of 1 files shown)


moosecanswim:
Reviewing

moosecanswim:
Works as expected.

Deployed normally with helm install, copied the sidecar certs to a file, applied the certs separately, helm installed with `--set sidecar.create_sidecar_secret=false`, waited for all pods to come up. setting `create_sidecar_secret results=true` results in it failing because of duplicate secrets

PR #503, opened 2020-03-10
Restructure

kaitmore:
This PR prepares for the big helm-charts refactor by moving all charts into three umbrella charts: fabric, sense, data. There are no code changes here, this is only a folder restructuring / consolidation of values.yaml files.

Changes
- Move all service charts into three umbrella charts
- Copy contents of greymatter.yaml into values.yaml for each parent chart and remove greymatter.yaml
- remove greymatter chart

To Test:
Parent charts can't be installed together yet because of secret conflicts. Run one, check that pods are up, then bring it down with `helm del --purge gm` before installing the next.

```
helm dep up fabric
helm install fabric -f greymatter-secrets.yaml -f credentials.yaml --set=global.environment=kubernetes --name gm
```

Note on sense: catalog will not come up because it's waiting for data/control, this is ok for now
```
helm dep up sense
helm install sense -f greymatter-secrets.yaml -f credentials.yaml --set=global.environment=kubernetes --name gm
```

```
helm dep up data
helm install data -f greymatter-secrets.yaml -f credentials.yaml --set=global.environment=kubernetes --name gm
```


moosecanswim:
LGTM

PR #507, opened 2020-03-11
Helm 3

zoemccormick:

This change fixes the restructure branch to use helm 3. There are a few changes - requirements.yaml is no longer necessary as the dependencies are now stored in the Chart.yaml files.  `serviceName` needed to be added to each statefulSet and the gm-control-api init script was removed because of a recursion issue.

<br/><br/>

No specific changes other than what we've already done for the restructure.

<br/><br/>

I added a bit of documentation on how to install helm 3 in https://github.com/DecipherNow/helm-charts/blob/helm-3/docs/Deploy%20with%20Minikube.md and removed the need to install Tiller. (but I assume these docs are going to be completely replaced after the rest of these changes anyway)

<br/><br/>

To use helm 3:

```
curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3
chmod 700 get_helm.sh
./get_helm.sh
```

`helm dep up` works the same as always.

To test any of the restructured charts:

```
helm dep up fabric
helm install gm-fabric fabric -f greymatter-secrets.yaml -f credentials.yaml --set=global.environment=kubernetes
```

will install fabric chart with name `gm-fabric`. To bring it down run `helm uninstall gm-fabric`.

<br/><br/>


data/jwt-gov/values.yaml
```diff
@@ -10,6 +10,14 @@ global:
...
   jwt:
     version: 1.0.0
+    sidecar:
+      version: latest
+  sidecar:
+    version: latest
+  spire:
+    enabled: false
+  mesh_tls:
+    use_provided_certs: true
```
kaitmore: can't wait to get rid of all this config

moosecanswim:
Looks good to me.  One thing to note and maybe we should add this to the contributor/ developer docs is that if a build fails then the resources it deploys will not be removed when doing a `helm uninstall`.  I had an install fail and had to remove the secrets and config maps manually before running again.

commands to remove:
```
kc delete secrets $(kc get secrets | awk '{ if($2=="Opaque") print $1 }') 
kc delete configmap $(kc get configmap | awk '{print $1}')
kc delete pvc $(kc get pvc | awk `{print $1}')
```
 the secrets removal is crude but most of our secrets were called typed opaque so it worked 🤷‍♂ 

All the following worked as expected:
`helm dep up fabric`
`helm install gm-fabric fabric -f credentials.yaml -f greymatter-secrets.yaml --set=global.environment=kubernetes`
`helm uninstall gm-fabric`

`helm dep up sense`
`helm install gm-data data -f greymatter-secrets.yaml -f credentials.yaml --set=global.environment=kubernetes`
`helm uninstall gm-data`

`helm dep up data`
`helm install gm-sense sense -f greymatter-secrets.yaml -f credentials.yaml --set=global.environment=kubernetes`
`helm uninstall gm-sense`

moosecanswim:
@zoemccormick @kaitmore I was just going back to the ci and noticed the v1 -> v2 change broke the ci.  https://circleci.com/gh/DecipherNow/helm-charts/766

According to this [post](https://github.com/helm/helm/issues/6753#issuecomment-545244626) it seems like this is a helm issue.  tomorrow ill work on the docker image we use to package charts to include helm3.

kaitmore:
@moosecanswim Ah good catch!! Thanks for working on the docker image.

PR #527, opened 2020-03-24
Make make

moosecanswim:
Makefiles were created for
- secrets
- voyager
- edge
- data
- fabric
- sense

Each make file has targets that will install the sub charts, the whole chart, and remove them.

I have tested each target on a minikube instance while assuming a helm3 install.

All makefiles check to see if the waiter sa is already in place and will add a flag to install that if it is not present.  When installing sub charts it is necessary to create the waiter sa manually before hand, whereas the parent charts have a template that will create the sa.

The make files are located under each of their respective folders to for secrets you will need to cd into secrets and run `make secrets`.


secrets/Makefile
```diff
@@ -0,0 +1,21 @@
+.PHONY: credentials
+credentials:
+	./../ci/scripts/build-credentials.sh
+
+copy-credentials:
+	@if [ -f ../credentials.yaml ] && [ ! -f ./credentials.yaml ]; then \
+		echo "No credentials in secrets. Coppying files from .."; \
+		cp ../credentials.yaml .; \
+	else \
+		echo "Double check that you have a valid credentials.yaml or we cant pull any images";\
```
kaitmore:
```suggestion
		echo "Double check that you have a valid credentials.yaml in the project root or we can't pull any images";\
```

secrets/Makefile
```diff
@@ -0,0 +1,21 @@
+.PHONY: credentials
+credentials:
+	./../ci/scripts/build-credentials.sh
+
+copy-credentials:
+	@if [ -f ../credentials.yaml ] && [ ! -f ./credentials.yaml ]; then \
+		echo "No credentials in secrets. Coppying files from .."; \
```
kaitmore:
```suggestion
		echo "No credentials in secrets. Copying files from .."; \
```

edge/Makefile
```diff
@@ -0,0 +1,14 @@
+WAITERSACHECK=$(shell kubectl get sa waiter-sa | tail -n +2 | awk '{if ($$1=="waiter-sa") print "present"}')
+ifeq (${WAITERSACHECK},present)
+	FLAG=--set=global.waiter.service_account.create=false
+else
+	FLAG=--set=global.waiter.service_account.create=true
+endif
```
kaitmore: Edge doesn't have a waiter service account so I think we can get rid of this

data/Makefile
```diff
@@ -0,0 +1,31 @@
...
+gm-data:
+	helm install data gm-data --set=global.environment=kubernetes $(FLAG)
+
+.PHONY: jwt
+jwt:
+	helm install data jwt --set=global.environment=kubernetes $(FLAG)
+
+# .PHONY: jwt
+# jwt-gov:
+# 	helm install data jwt-gov --set=global.environment=kubernetes
```
kaitmore: Why comment this out?

data/README.md
```diff
@@ -49,3 +49,13 @@ Alternatively, a YAML file that specifies the values for the above parameters ca
 ```sh
 helm install sense --name <my-release> -f custom.yaml
 ```
+
+## Makefile
+
+| command      | description                        | comments |
+| ------------ | ---------------------------------- | -------- |
+| gm-data      | install standalone data            |          |
+| jwt          | install standalone jwt             |          |
```
kaitmore: Add jwt-gov?

moosecanswim:
@zoemccormick whoops totally missed the edge readme.  I talked with @kaitmore and at this point people above want this split into seperate sections... but the way i set it up i think i can just `include` the individual make files, add a target and it should be good to do.

moosecanswim:
Thanks for the feedback.  I've updated the pr to make those changes.  

Oh i misunderstood Zoe's comment.  Gotcha.  Will do!

moosecanswim:
@zoemccormick annd @kaitmore I think this is ready to go!  Chris asked me to migrate to using k3d which posed some issues but everything should be ready to do.

ci/scripts/minikube.sh
```diff
@@ -17,18 +17,21 @@ if [ $LC -ge 4 ]; then
...
+#     echo "Installing local charts"
+#     REPO=greymatter
+#     helm dep up greymatter
+#     # helm dep up edge fabric sense data
+#     # helm install edge
+#     # helm instal
+# fi
+
+# helm install $REPO -f greymatter.yaml -f greymatter-secrets.yaml -f credentials.yaml --set global.environment=kubernetes -n gm-deploy
+# ./ci/scripts/show-voyager.sh
```
kaitmore, line 36: Should this whole block be commented out?

ci/scripts/k3d.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# install k3d 1.6.0
+curl -s https://raw.githubusercontent.com/rancher/k3d/master/install.sh | TAG=v1.6.0 bash
```
kaitmore, line 4: Should this script actually install k3d for you? I feel like we should just have this as a prerequisite in the "Deploy with k3d" doc, that way the script can still be used to start up the cluster after the first install.

moosecanswim: It will work after an install.  It just responds back that it already has k3d and the version.

ci/scripts/minikube.sh
moosecanswim: I can delete it.  The minikube script did allot of things including starting up minikube, adding voyager, and installing greymatter.  I've separated out these into their own make targets to support multiple deployment environments.

ci/scripts/k3d.sh
kaitmore: Ahh ok sweet!

whenhellfreezes:
Saw an error for Not Found "waiter-sa". May want to figure out what caused that or is that intended behaviour? Also I think we should just depreciate minikube and that should be reflected in the docs.

moosecanswim:
@whenhellfreezes 
The error is expected.  Basically each time make attempts to install a service that expects the waiter-sa it checks to see if the service account exists.  if it does not then you get that error but it adds a flag to the helm install so it installs the waiter sa.

Saw an error for Not Found "waiter-sa". May want to figure out what caused that or is that intended behaviour? Also I think we should just depreciate minikube and that should be reflected in the docs.

kaitmore:
@moosecanswim When I run `make destroy`:

```
make destroy
Makefile:22: *** missing separator.  Stop.
```

moosecanswim:
@kaitmore whoops, i was working on another feature then decided against that and didnt remove all of the commits.  Just pushed a commit to fix that.

PR #532, opened 2020-03-26
Mongo userz

moosecanswim:
changes the mongo init config map into a secret.

to test start mesh in either k3d or minikube and make sure dashboard is up and running with all core services green.

run `kc logs data-mongo-0` to see logs of mongo.  You can also do this for the internal mongo.


data/templates/mongo-init.yaml
```diff
@@ -0,0 +1,21 @@
...
+  gmdata.js: |-
+  {{- with .Values.mongo.credentials }}
+    db.auth({{ quote .root_username }}, {{ quote .root_password }})
+    db = db.getSiblingDB({{ quote .database }})
+    db.createUser({
+        user: {{ quote .gmdata_username }},
+        pwd: {{ quote .gmdata_password }},
+        roles: [
+        {
+            role: {{ quote .gmdata_role }},
```
chrisbsmith: @moosecanswim I updated this file.  The GM Data user does not need root rights to the admin db.  Instead. we should allow this to be configurable.  I've updated the code and given the `readWrite` default.

data/values.yaml
```diff
@@ -250,8 +250,9 @@ mongo:
     root_username: 'mongo'
     root_password: 'mongo'
     database: 'gmdata'
-    gmdata_username: 'gmdata-user'
-    gmdata_password: 'mongopassword'
+    gmdata_username: 'gmdata'
```
chrisbsmith: We should also not use the same username and password for the GM Data user as the root user, b/c we'll mask any potential permissions issues.

moosecanswim:
Can do

moosecanswim:
Works as expected and i incremented the chart version

PR #533, opened 2020-03-27
Read permission

hiromis:

Updates required to make the chart OpenShift4 and CodeReady compatible.
Also, made `openshift` as the default all across (some were set to kubernetes).

New local.yaml with limited resources and updated route.

All the notes can be found here: https://notes.deciphernow.com/t/codeready-container-on-mac/813

Once the charts become stable and ready, I can also add an actual documentation in this repo.

See notes above.


local.yaml
```diff
@@ -0,0 +1,233 @@
+global:
```
kaitmore: Could we rename this file something like `codeready.yaml` and put it in a new directory called `overrides`? We will probably end up having a few different values files for various scenarios, and it would be nice to clean up the root directory. fyi @chrisbsmith

hiromis: Done!

hiromis:
> You might want to update this in your instructions,

Right, I think I started to update it at some point... Completed the sentence 😅

> Double check the credentials.yaml and make sure it’s pointing to docker-dev.production.deciphernow.com if you are using dev only images. Otherwise, it’s okay.

PR #541, opened 2020-04-01
Add Spire

zoemccormick:
This branch adds spire support using the spire registrar service.  Any pods with the label `gmproxy` will have an identity created and accessible only to that pod. 

To test in minikube - follow the instruction in [this doc](https://github.com/DecipherNow/helm-charts/blob/jwt-spire-registrar/docs/spire/Minikube%20SPIRE.md).   Step 3 is done by default in this branch currently, so all of configurations should be good to go.

To see the registration entries, exec into the spire server, `kubectl exec -it server-0 /bin/sh -n spire -c server` and run `/opt/spire/bin/spire-server entry show -registrationUDSPath /run/spire/socket/registration.sock` There should be an entry that looks like the following for every pod with label "gmproxy" 

```
Entry ID      : 0016c1ce-c826-4ca2-aa32-16e5f0d5a66a
SPIFFE ID     : spiffe://quickstart.greymatter.io/slo
Parent ID     : spiffe://quickstart.greymatter.io/k8s-workload-registrar/quickstart/node
TTL           : default
Selector      : k8s:ns:default
Selector      : k8s:pod-name:slo-58585dc6c5-pcf5c
```

To verify that the sidecars are fetching the certs, exec into any sidecar and run `curl localhost:8001/certs`, run `curl localhost:8001/config_dump` to see the `dynamic_active_secrets` at the bottom.


hiromis:
I am just noting what I came across when I tried to run this locally.

The very first time, everything went smoothly until I ran `helm install sense sense ...` - which resulted in timeout due to resource shortage.

The second time around, I was seeing this error `cannot re-use a name that is still in use` when trying to install secrets. Tried suggestions in these to do clean install, but no luck:

- https://github.com/helm/helm/issues/4174
- https://stackoverflow.com/questions/48077931/delete-all-the-contents-from-a-kubernetes-node

Per @kaitmore 's suggestion, maybe the way forward would be to use k3d instead of minikube.
Since there are some issues to be ironed out for this PR, I am setting this aside for the time being.

PR #550, opened 2020-04-06
add in spire makefile

moosecanswim:
This PR adds a makefile for spire helm releases and updates the top level Makefile to install spire with the rest of the Grey Matter core components.

To test you can start up a k3d cluster and then run `make install`


moosecanswim:
Thanks! I'll make that change.

PR #593, opened 2020-05-04
User DN etc are now configurable

### LINKED ISSUE #580: [release-2.2] quick start DN is hard coded in the spire configuration
While the helm-charts use the quickstart cert for easy introduction to Grey Matter, the quickstart DN has been hard coded into the envoy rbac filters in the listeners.  Since it's hard coded, it is not configurable and will break when using real certs.

These should be configurable from the `global.yaml` file

https://github.com/DecipherNow/helm-charts/blob/a87d509c3cff036fa79bf3cddcc1b86ab8397ef1/fabric/control-api/json/services/listener.json#L47

hiromis:
Closes #580 

As you see in fabric/control-api/values.yaml, these were made configurable:
```
  rbac:
    admin:
      header_name: user_dn
      match_type: exact_match
      header_value: CN=quickstart,OU=Engineering,O=Decipher Technology Studios,L=Alexandria,ST=Virginia,C=US
```

If user_dn needs to be overridden, you can add something like the following to global.yaml:
```
  rbac:
    admin:
      header_value: new-value
```

How to test:

I removed lines for spire_volume_mount (control-api/templates/control-api.yaml) as it is to be dealt in another issue. But for references, the error looked like:
```
helm template fabric/control-api -f global.yaml --debug

install.go:158: [debug] Original chart version: ""
install.go:175: [debug] CHART PATH: /Users/hiromi/git/helm-charts/fabric/control-api

Error: template: control-api/templates/control-api.yaml:114:14: executing "control-api/templates/control-api.yaml" at <include "spire_volume_mount" .>: error calling include: template: no template "spire_volume_mount" associated with template "gotpl"
helm.go:75: [debug] template: control-api/templates/control-api.yaml:114:14: executing "control-api/templates/control-api.yaml" at <include "spire_volume_mount" .>: error calling include: template: no template "spire_volume_mount" associated with template "gotpl"
```

After that, run this command:
```
 helm template fabric/control-api -f global.yaml --debug > output
```

and make sure all listener.json has `new-value` instead of `CN=quickstart,OU=Engineering,O=Decipher Technology Studios,L=Alexandria,ST=Virginia,C=US`

### DIFF
--- a/fabric/control-api/json/services/listener.json
+++ b/fabric/control-api/json/services/listener.json
@@ -60,7 +60,7 @@
                       "principal_name": {
                         "safe_regex": {
                           "google_re2": {},
-                          "regex": "spiffe:\/\/quickstart.greymatter.io\/(.+)"
+                          "regex": "spiffe:\/\/{{ .Values.global.spire.trust_domain }}\/(.+)"
                         }
                       }
                     }
--- a/fabric/control-api/json/services/listener.json
+++ b/fabric/control-api/json/services/listener.json
@@ -43,8 +43,8 @@
                   }],
                   "principals": [{
                       "header": {
-                          "name": "user_dn",
-                          "exact_match": "CN=quickstart,OU=Engineering,O=Decipher Technology Studios,L=Alexandria,ST=Virginia,C=US"
+                          "name": "{{ .Values.global.rbac.admin.header_name }}",
+                          "{{ .Values.global.rbac.admin.match_type }}": "{{ .Values.global.rbac.admin.header_value }}"
                       }
                 }]
               },
@@ -66,7 +66,7 @@
                     }
                   }
                 ]
-              },    
+              },
               "product-viewer": {
                   "permissions": [{
                       "header": {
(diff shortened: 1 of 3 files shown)


fabric/control-api/json/services/listener.json
```diff
@@ -43,8 +43,8 @@
                   }],
                   "principals": [{
                       "header": {
-                          "name": "user_dn",
-                          "exact_match": "CN=quickstart,OU=Engineering,O=Decipher Technology Studios,L=Alexandria,ST=Virginia,C=US"
+                          "name": "{{ .Values.global.rbac.admin.header_name }}",
+                          "{{ .Values.global.rbac.admin.match_type }}": "{{ .Values.global.rbac.admin.header_value }}"
```
chrisbsmith, line 47: It's too far from this code, but can you also update line 63?  Specifically where it says

``` 
"regex": "spiffe:\/\/quickstart.greymatter.io\/(.+)"
```

hiromis: Just to double check - would you like that to be changed to `.Values.global.spire.trust_domain`?

hiromis:
Sorry about that. I work on projects where trailing whitespaces will cause issues so my editor removes trailing spaces on save action. I will be mindful in future, but if I do forget, there is this thing:
Screen Shot 2020-05-04 at 2 39 10 PM

chrisbsmith:
Thanks for that tidbit @hiromis

fabric/control-api/json/services/listener.json
chrisbsmith: Yes, That makes the most sense

hiromis: Done!

PR #598, opened 2020-05-04
If spire is disabled, the spire charts should not be installed 

### LINKED ISSUE #588: [release-2.2] If spire is disabled, the spire charts should not be installed
I disabled spire to test the cert deployment and the spire agent and server were still deployed. I wouldn't expect these to be deployed. The Makefile should detect this and not deploy spire

hiromis:
Closes #588 

Added a question regarding whether or not to install spire. When you run `make install`, the only way spire will be installed is for you to type `y` - everything else is considered as no.

Also the change for `docs/Deploy with K3d.md` is removing a strange character:
Screen Shot 2020-05-04 at 6 34 30 PM

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,9 +53,9 @@ check-secrets:
 install-spire:
 	@read -p "Would you like to install Spire? [y/N] " response; \
 	if [ "$$response" = "y" ]; then \
-        echo "Installing spire"; \
+		echo "Installing spire"; \
 		(cd spire && make spire); \
-    fi
+	fi
 
 .PHONY: install
 install: dev-dep check-secrets install-spire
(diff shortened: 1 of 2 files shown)


moosecanswim:
@hiromis i tested this and realized that we do need the chart.x files for the `helm dep up .` that happens in the `package spire` target.  

i ran a `make destroy` and `make k3d` for a clean cluster and `make install` (answer yes to install spire) could not get a working dashboard.

hiromis:
@moosecanswim - understood! I reverted the last two commits, and running:
```
make destroy
make k3d 
export KUBECONFIG="$(k3d get-kubeconfig --name='greymatter')" 
make install
```
seems to work fine on my machine.

hiromis:
Hi @moosecanswim! The requested changes have been made. Could you take a look?

PR #609, opened 2020-05-08
use global.yaml spire.enabled value to decide to install spire or not

moosecanswim:

Updates the install-spire makefile target to use the `global.spire.enabled` value to determine if helm should install spire.

None

None needed

Change the `global.spire.enabled=true` and run `make install-spire` To verify it installs use `kubectl get pods --all-namespaces`.

Follow the same with false to ensure spire is not installed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,18 +45,14 @@ dev-dep: clean
 .PHONY: check-secrets
 check-secrets:
 	$(eval SECRET_CHECK=$(shell helm ls | grep secrets | awk '{if ($$1 == "secrets") print "present"; else print "not-present"}'))
-	echo $(SECRET_CHECK)
 	if [[ "$(SECRET_CHECK)" != "present" ]]; then \
-		echo "present" ;\
 		(make secrets);\
 	fi
 
 .PHONY: install-spire
 install-spire:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -53,8 +53,9 @@ check-secrets:
 
 .PHONY: install-spire
 install-spire:
-	@read -p "Would you like to install Spire? [y/N] " response; \
-	if [ "$$response" = "y" ]; then \
+	$(eval IS=$(shell grep -A3 'spire:' global.yaml | grep enabled: | awk '{print $$2}'))
```
chrisbsmith, line 56: Rather than running an `eval` statement, let's use this instead
```suggestion
IS=$(cat global.yaml| grep -A3 'spire:'| grep enabled: | awk '{print $2}')
```

Makefile
```diff
@@ -53,8 +53,9 @@ check-secrets:
...
+	echo $(IS)
```
chrisbsmith, line 57: Let's remove this - unless there's a way to have a debug mode in Makefiles. But then you'd need to add the debug check around this.

Makefile
moosecanswim: That wont work.  because make executes commands line to line as independent scripts.  this was the only way i could find to create envars that can be used throughout targets.

moosecanswim: refering to `eval` and `shell`

PR #627, opened 2020-05-14
Add the ability to have a global cert and a unique edge ingress cert

chrisbsmith:

There are times when we may want to use the global cert option from #597 but also need to provide the edge ingress with a unique key. This PR updates the secrets edge ingress certs to allow for a unique cert. 

2. What **changes to global.yaml** are required?

none

the configuration.md has been updated

`cd secrets` and run `make template-secrets` with various configurations:
- `global.global_certs.enabled` = `true`/`false`
- `secrets/values.yaml`, `edge.certificates.ingress.unique` = `true`/`false`


chrisbsmith:
Great catch @zoemccormick .  Your second option may be the simpler option. In that case, the edge-ingress cert will just always be created, regardless of the settings.  As long as the `unique` cert still allows for a different cert to be populated when global is set to true.

Let me know if we need to discuss this further, but I think you have it!

chrisbsmith:
@zoemccormick - excellent work with this!

PR #636, opened 2020-05-19
update fabric chart to have just one jwt-gov

moosecanswim:

Modify ci script to package jwt (regular for internal) and jwt-gov in fabric-gov chart

none

none

ci passed for helm-staging and deployment worked as expected.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -96,7 +96,6 @@ jobs:
             cp $(pwd)/fabric/Chart.yaml $(pwd)/fabric/Chart.yaml.orig
             echo "$(awk 'NR==1,/name: jwt/{sub(/name: jwt/, "name: jwt-gov")} 1' fabric/Chart.yaml)" > fabric/Chart.yaml
             sed -i.bak-gov -e "s|name: fabric|name: fabric-gov|" -e "s|Grey Matter Fabric|Grey Matter Fabric-gov|" $(pwd)/fabric/Chart.yaml
-            cat "$(pwd)/fabric/Chart.yaml"
             docker run --rm -e INPUT_NEXUS_URL=$NEXUS_URL -e INPUT_NEXUS_USER=$NEXUS_USER -e INPUT_NEXUS_PASS=$NEXUS_PASSWORD -v $(pwd):/helm deciphernow/helm-packager:v3.1.1 /helm/fabric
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -93,10 +93,12 @@ jobs:
       - run:
           name: Package and Publish Fabric-gov
           command: |
-            sed -i.bak-gov -e "s|name: jwt|name: jwt-gov|" -e "s|name: fabric|name: fabric-gov|" -e "s|Grey Matter Fabric|Grey Matter Fabric-gov|" $(pwd)/fabric/Chart.yaml
+            cp $(pwd)/fabric/Chart.yaml $(pwd)/fabric/Chart.yaml.orig
+            echo "$(awk 'NR==1,/name: jwt/{sub(/name: jwt/, "name: jwt-gov")} 1' fabric/Chart.yaml)" > fabric/Chart.yaml
+            sed -i.bak-gov -e "s|name: fabric|name: fabric-gov|" -e "s|Grey Matter Fabric|Grey Matter Fabric-gov|" $(pwd)/fabric/Chart.yaml
             cat "$(pwd)/fabric/Chart.yaml"
```
chrisbsmith, line 99: You don't need to cat the output anymore

.circleci/config.yml
```diff
@@ -93,10 +93,12 @@ jobs:
...
             docker run --rm -e INPUT_NEXUS_URL=$NEXUS_URL -e INPUT_NEXUS_USER=$NEXUS_USER -e INPUT_NEXUS_PASS=$NEXUS_PASSWORD -v $(pwd):/helm deciphernow/helm-packager:v3.1.1 /helm/fabric
-            mv $(pwd)/fabric/Chart.yaml.bak-gov $(pwd)/fabric/Chart.yaml
+            mv $(pwd)/fabric/Chart.yaml.orig $(pwd)/fabric/Chart.yaml
```
chrisbsmith, line 101: I assume this works because you're running the docker packager one line before you overwrite the  chart.yaml file?

moosecanswim: Yeah, basically we're changing the dependencies in Chart.yaml then we need to put it back the way it was since "checkout" is not a `git reset origin/...`.  

from one stage to another this shouldn't really matter but if we ever come back and package another fabric then this will ensure the chart.yaml is in its original state

PR #653, opened 2020-06-03
update secrets to allow in rds certs over a global cert and include a…

moosecanswim:
… check in secrets to ensure rds.enables values match

<br/><br/>

adds rds.enabled under posgres in the secrets/values.yaml file

none needed.  there is however a check to make sure that if the sense rds enabled matches the secrets rds enabled values

*start*
with a clean k3d cluster
`make credentials`
set global_certs true in global.yaml
`make secrets`
check secrets with `kc get secrets | grep postgres` ensure postgres-ssl-certs is not listed
`make remove secrets`

*Make sure value validator works*
set `postgres.rds.enabled` to true in `secrets/values.yaml` and leave it false in `sense/values.yaml`
`make secrets`
The make target should exit with an error 2

*Create special for rds*
set `postgres.rds.enabled` to true in `secrets/values.yaml` and `sense/values.yaml`
in `secrets/values.yaml` set `postgres .ssl.certificates.ca` to "apples"
`make secrets`
check secrets with `kc get secrets | grep postgres` ensure postgres-ssl-certs is listed
check to ensure your cert was processed correctly with `kc get secrets postgres-ssl-certs -o yaml`. Copy the ca value and base6d decode it.  you should get apples.

### DIFF
--- a/secrets/Makefile
+++ b/secrets/Makefile
@@ -8,39 +8,30 @@ pwd_name := $(notdir $(PWD))
 
 .PHONY: credentials
 credentials:
-	@if [ "$(pwd_name)" == "helm-charts" ]; then \
-		./ci/scripts/build-credentials.sh; \
-	elif [ "$(pwd_name)" == "secrets" ]; then \
-		./../ci/scripts/build-credentials.sh; \
-	else \
-		echo "Make sure you are running this from helm-charts or secrets makefile"; \
-	fi
+	./../ci/scripts/build-credentials.sh
 
 copy-credentials:
-	@if [ "$(pwd_name)" == "helm-charts" ]; then \
-		cp ./credentials.yaml ./secrets/credentials.yaml; \
-	elif [ "$(pwd_name)" == "secrets" ]; then \
-		cp ../credentials.yaml .; \
-	else \
-		echo "Make sure you are running this from helm-charts or secrets makefile"; \
-	fi
+	cp ../credentials.yaml .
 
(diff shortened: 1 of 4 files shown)


secrets/values.yaml
```diff
@@ -1302,7 +1304,7 @@ jwt:
         PK57iQEQz3QrshXB5nrvhAjKfiESo++zFmkZUAqgTjrox2EJUjtHFx40YbFe1Vrf
         N69uc/UauO74PKYWRWU=
         -----END PUBLIC KEY-----
-      jwt.api_key: dm9sY2Fuby1lcGlkZW1pYy10d2VsZnRoLXRhbWFsZQ==
+      # jwt.api_key: dm9sY2Fuby1lcGlkZW1pYy10d2VsZnRoLXRhbWFsZQ==
```
chrisbsmith, line 1307: Is this included at a higher level?  That's why it's safe to comment out?

moosecanswim: oh good catch.  thats from another thing i was working on

secrets/Makefile
```diff
@@ -8,39 +8,30 @@ pwd_name := $(notdir $(PWD))
...
+secrets: copy-credentials values-validation
+	helm install secrets . -f credentials.yaml -f ../global.yaml
+	rm credentials.yaml
 
 .PHONY: remove-secrets
 remove-secrets:
 	helm uninstall secrets
+
+
+values-validation:
```
chrisbsmith, line 30: I like this! 💯

PR #663, opened 2020-06-10
remove rbac config from listener without spire

### LINKED ISSUE #644: listener spire cruft created even when spire is not enabled
When spire is disabled, the template does not add the `envoy.rbac` filter to the active filters, yet it leaves the below http filter information in the listener. This is not needed and is extra cruft that can leave people astray when debugging.

When spire is disabled, this should also be removed.

https://github.com/DecipherNow/helm-charts/blob/bb0f3a20679ddd90cf375ca8a1464842b4eb9e5a/fabric/control-api/json/services/listener.json#L35-L83

zoemccormick:
quick fix removes the rbac config when rbac is not enabled without spire

closes #644 

To test, set global.spire.enabled to false and 

```
helm dep up fabric
helm dep up edge
helm install fabric fabric -f global.yaml
helm install edge edge -f global.yaml
```

Once it runs, `greymatter list listener` -> the only listener with envoy_rbac configuration in its http_filters config should be `edge-listener`.


fabric/control-api/json/special/internal-cluster-catalog-data-internal.json
```diff
@@ -5,8 +5,8 @@
   "require_tls": true,
   "secret": {
     "secret_key": "secret-data-internal-secret",
-    "secret_name": "spiffe://quickstart.greymatter.io/catalog",
-    "secret_validation_name": "spiffe://quickstart.greymatter.io",
+    "secret_name": "spiffe://{{ .Values.global.spire.trust_domain}}/catalog",
```
chrisbsmith, line 8: Does this file still get populated regardless of the use of spire or not?  If so, we should probably set a default value for these in the case of a non-spire install when `.Values.global.spire.trust_domain` may not be defined.

zoemccormick: @chrisbsmith  there should be a default for that value in every values.yaml file (quickstart.greymatter.io), and these "internal" clusters should not be applied in a non-spire install. In general we should probably revamp all of these to make the mesh configs better for non spire (to still use internal clusters/egress listeners and set ssl_configs)

chrisbsmith: Sounds good, thanks @zoemccormick

PR #681, opened 2020-07-02
WIP: configure prometheus ALS in gm-control

### LINKED ISSUE #680: add option to prometheus scrape gm-control
gm-control's built in access logger can be configured to output its built-in ALS stats to a prometheus endpoint. This should most likely be a toggleable option.

critterjohnson:

resolves #680
<br/><br/>

Added global.control.prometheus_port to configure a port for the prometheus client to listen on.
<br/><br/>

prometheus_port
<br/><br/>

To test:

`helm install` or `helm upgrade` to these charts. Make sure to set `gm_control_stats_addr`'s type to `value` before you install/upgrade. Also make sure you're using the latest version of gm-control (might require building locally and importing the images into your cluster). 
<br/><br/>

### DIFF
--- a/fabric/control/values.yaml
+++ b/fabric/control/values.yaml
@@ -97,13 +97,13 @@ control:
     gm_control_kubernetes_namespaces:
       type: 'value'
       value: '{{ include "control.namespaces" . }}'
+    # The following values are not applied. Change the 'type' to 'value' to use these variables.
     gm_control_stats_backends:
-        type: 'value'
+        type: 'null'
         value: 'prometheus'
     gm_control_stats_addr:
-      type: 'value'
+      type: 'null'
       value: '0.0.0.0:{{ .Values.global.control.prometheus_port }}/prometheus'
-    # The following values are not applied. Change the 'type' to 'value' to use these variables.
     gm_control_diff_dry_run:
       type: 'null'
       value: 'true'
--- a/fabric/control/values.yaml
+++ b/fabric/control/values.yaml
@@ -18,6 +18,8 @@ global:
     port: 50000
     # The label Control uses to find pods to include in the mesh
     cluster_label: greymatter.io/control
+    # port for prometheus to listen on
+    prometheus_port: 9102
     # Comma delimited string of namespaces for control to monitor. Also used by prometheus. 
     additional_namespaces:
   # Configures the init container used to wait on various deployments to be ready
@@ -95,6 +97,12 @@ control:
     gm_control_kubernetes_namespaces:
       type: 'value'
       value: '{{ include "control.namespaces" . }}'
+    gm_control_stats_backends:
+        type: 'value'
+        value: 'prometheus'
+    gm_control_stats_addr:
+      type: 'value'
+      value: '0.0.0.0:{{ .Values.global.control.prometheus_port }}/prometheus'
     # The following values are not applied. Change the 'type' to 'value' to use these variables.
     gm_control_diff_dry_run:
       type: 'null'
--- a/fabric/values.yaml
+++ b/fabric/values.yaml
@@ -20,6 +20,8 @@ global:
     port: 50000
     # The label Control uses to find pods to include in the mesh
     cluster_label: greymatter.io/control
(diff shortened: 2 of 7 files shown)


fabric/control/values.yaml
```diff
@@ -96,6 +98,12 @@ control:
       type: 'value'
       value: '{{ include "control.namespaces" . }}'
     # The following values are not applied. Change the 'type' to 'value' to use these variables.
+    gm_control_stats_backends:
+        type: 'null'
+        value: 'prometheus'
+    gm_control_stats_addr:
+      type: 'null'
```
chrisbsmith, line 105: Your instructions say to set this to `value` but it's left as `null` here.  Can this be set to `type` to prevent manual interaction?

fabric/values.yaml
```diff
@@ -76,8 +78,8 @@ control:
     # Number of replicas for the deployment
     replicas: 1
     # Version of control
-    version: 1.4.2
-    image: 'docker.production.deciphernow.com/deciphernow/gm-control:{{ $.Values.control.version }}'
+    version: latest
```
chrisbsmith, line 81: we don't deploy `latest`. These need to be pinned to a version

justincely: see my comment.  this is still WIP while a release happens and we talk about the added burden to prometheus

chrisbsmith:
@justincely coming back around to this one.  Should we work this PR a little more to release it with v1.3?

justincely:
@chrisbsmith we certainly can, though I think that's more your call than mine at this point.  The stats that this would output is [in our docs](https://docs.greymatter.io/v/1.3-beta/setup/fabric/gm-control/als_stats), maybe you could take a look and see if you'd like this in the charts at all?

chrisbsmith:
@justincely I don't understand the benefit to this PR well enough to make a priority decision. Can you explain a little more why this is useful?

justincely:
yeah, essentially `gm-control` always serves as the ALS (Access Log Service) in the mesh.  Every sidecar is auto configured to send off a log when a call comes through, and they send it to gm-control.  `gm-control` doesn't typically do much with these access logs, though it can do something called "leaderboard logging" that it can dump into it's own logs.

This PR lets the charts turn on the ability for gm-control to dump the internal stats that it collects from these logs out to prometheus.  The docs that I linked show what these stats are, but my initial thought is that they're largely duplicates of what we get from observables.  

So I think you should just look at the stats from that doc I linked and see if you want any of those in prometheus.  I have a feeling we'll never use it if we don't make it easy in the helm charts, but that it will be a decent increase in prometheus usage.

chrisbsmith:
@justincely is Prometheus really the best output for these logs? These seem like they would be better off dumping to STDOUT and being picked up by Splunk or so.  

I'd also like to understand how much data it's adding to Prometheus. We run with a finite amount of disk space, as do our customers, do it'd be good to know how this will increase the rate of disk consumption in Prometheus.

Also, do we bring a graph with this new data set?  It's great to dump it to Prometheus, but if we don't provide an easy way to view it, then it doesn't do us much good.

justincely:
a graph for the dashboard?  Or in grafana?  Either way, the answer is no.  This is a feature we "found" from the old code we inherited and just had an intern re-enable.  That's basically all we really know about it.

At this point, with all the unknowns, i'm tempted to bring it in as an option.  We won't know what it is, how much data it adds, etc until we use it more and we won't use it more unless it's an option in the helm charts.  Still, I think these helm charts lie in your ownership, so if they should be reserved for "production ready" stuff then i'm fine to drop this until we spend the time to understand it.

tilleryd:
This PR fell behind and I attempted to merge in release-2.4. However, due to the number of conflicts I opted to recreate this  against a new branch off of release-2.4. This doesn't need to be targeted for merge into any release branch until we understand the benefits/impacts of turning this on. Thus, I'm going to close this PR and put a product developer (or two) on this effort for the 2021 Sprint 03. using the new `gm-control-prom-stats` branch. @chrisbsmith All of your PR comments are addressed in this new branch.

PR #683, opened 2020-07-06
651 egress

### LINKED ISSUE #651: Outbround traffic from catalog and data-internal not using egress proxy
There is a bit of misconfiguration for the `catalog -> data-internal` and `data-internal to internal-jwt` communication. Currently, both communication paths are set to use the ingress port `10808` rather than the egress port of `10909`.

This needs to be updated on the service configurations, but also - and more importantly - in the mesh configuration objects. Specifically, the `/data` and `/jwt` routes will need to use the egress domains for the `catalog` and `data` services, respectfully.

zoemccormick:

<br/><br/>

closes #651 

This updates mesh configs and service env vars to use egress domains for internal routes in the non-spire install

<br/><br/>

None

Details:

This PR does a couple of things to update the egress routing:

- updates any service env vars making requests to other services in the mesh to go through their egress listener port (10909). These egress listeners/domains are [already created and applied](https://github.com/DecipherNow/helm-charts/blob/9623b425f72f896d4720187f1af0db4df84a8d60/fabric/control-api/config/bootstrap.sh#L58-L60) and just have not been used
- updates the internal routes (catalog -> internal-data -> internal-jwt) to set those routes on the egress domains so the request generated by the service above goes through the egress listener on 10909, which has the proper route set to point to the next service
- generates an internal-cluster with an ssl config for all internal routes and unique shared-rules objects for those clusters (the same way spiffe/spire does) now applied [here](https://github.com/DecipherNow/helm-charts/blob/2f76a6c904b3c04ee489e07b16496dcd46551691/fabric/control-api/config/bootstrap.sh#L109-L117). The reason for this:
    - current behavior reuses the `edge-to-serviceB-cluster` for any internal route from serviceA to serviceB - for example using `edge-to-data-internal-cluster` for catalog -> data-internal route [here](https://github.com/DecipherNow/helm-charts/blob/9623b425f72f896d4720187f1af0db4df84a8d60/fabric/control-api/json/special/route-catalog-to-internal-data.json#L15). This is technically fine based on the way the helm charts work and configure ssl, but logically not quite right.  It only works because all sidecars by default use the quickstart certs and mount them at the same place, so [this ssl_config](https://github.com/DecipherNow/helm-charts/blob/9623b425f72f896d4720187f1af0db4df84a8d60/fabric/control-api/json/edge/cluster.json#L23-L32) is the same as this [ssl_config](https://github.com/DecipherNow/helm-charts/blob/2f76a6c904b3c04ee489e07b16496dcd46551691/fabric/control-api/json/special/internal-cluster-catalog-data-internal.json#L20-L29) which is why reusing this edge cluster works. In any case where certs mounted into the catalog container are not named `server.crt`, `server.key` etc or mounted at the same place `/etc/proxy/tls/sidecar/` this logic fails, so i made this particular change mainly for clarity for people trying to adapt the helm charts to a more specific configuration

The first two points are directly aligned with the issue, and the third is more of an addition to help clean up the egress logic.

### DIFF
--- a/fabric/control-api/config/bootstrap.sh
+++ b/fabric/control-api/config/bootstrap.sh
@@ -105,17 +105,17 @@ done
 
 cd $MESH_CONFIG_DIR/special
 
-if $SPIRE_ENABLED; then
-    echo "Adding internal clusters"
-    for cl in $(ls internal-cluster-*.json); do
-        create_or_update "cluster" $cl
-    done
 
-    echo "Adding internal shared_rules"
-    for sr in $(ls internal-shared-rules-*.json); do
-        create_or_update "shared_rules" $sr
-    done
-fi
+echo "Adding egress clusters"
+for cl in $(ls internal-cluster-*.json); do
+    create_or_update "cluster" $cl
+done
+
+echo "Adding egress shared_rules"
+for sr in $(ls internal-shared-rules-*.json); do
+    create_or_update "shared_rules" $sr
+done
+
 
 echo "Adding additional Special Routes"
 for rte in $(ls route-*.json); do
--- a/data/gm-data/values.yaml
+++ b/data/gm-data/values.yaml
@@ -148,7 +148,7 @@ data:
     # Use data's egress listener port
     client_jwt_endpoint_port:
       type: value
-      value: '{{- if .Values.global.spire.enabled }}10909{{- else}}10808{{- end }}'
+      value: '10909'
     client_jwt_endpoint_prefix:
       type: value
       value: '/jwt'
--- a/data/values.yaml
+++ b/data/values.yaml
@@ -181,7 +181,7 @@ data:
         value: localhost
       client_jwt_endpoint_port:
         type: value
-        value: '{{- if .Values.global.spire.enabled }}10909{{- else}}10808{{- end }}'
+        value: '10909'
       client_jwt_endpoint_prefix:
         type: value
         value: /jwt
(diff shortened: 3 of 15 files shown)


fabric/control-api/config/bootstrap.sh
```diff
@@ -105,17 +105,17 @@ done
...
-    for cl in $(ls internal-cluster-*.json); do
-        create_or_update "cluster" $cl
-    done
 
-    echo "Adding internal shared_rules"
-    for sr in $(ls internal-shared-rules-*.json); do
-        create_or_update "shared_rules" $sr
-    done
-fi
+echo "Adding egress clusters"
```
chrisbsmith, line 109: In a non-SPIRE enabled environment, is this now poluting the mesh with a bunch objects we won't use?

chrisbsmith:
@zoemccormick there seems to be a lot that's going on in this PR to fix the egress port. Can you provide a little more details than what PR it's fixing?

zoemccormick:
@chrisbsmith yes of course updated in the comment, thanks!

chrisbsmith:
Thank you @zoemccormick !

chrisbsmith:
@zoemccormick thank you for the detailed update! I appreciate the changes you have made to create a use new clusters, routes and shared rules for internal traffic.

While I haven't tested this, I fully agree with the concepts and look forward to getting this into the code base.

PR #691, opened 2020-07-16
updating edge service type

### LINKED ISSUE #669: Edge Service is defined as a LoadBalancer
We are making quite the assumption by hard coding the edge service resource as a type `LoadBalancer`. Not only does this cost additional money in the cloud environment (spinning up a separate LB), it quite possibly circumvents the ingress configuration defined by the cluster.

In our documentation, we describe how to configure voyager or nginx-ingress to handle cluster ingress. These two tools manage all inbound traffic, yet when we specify the edge as a LoadBalancer type, it creates it's own ingress path. This should be removed in favor of a type `ClusterIP` (or left blank which defaults to `clusterIP`), as was done in `release-2.1`

https://github.com/DecipherNow/helm-charts/blob/79e1cf58d1c615b77a481e4da2d1000f750f898a/edge/templates/edge-service.yaml#L17

https://github.com/DecipherNow/helm-charts/blob/release-2.1/edge/templates/edge-service.yaml

rahug20:

Updated edge service type from Load Balancer to empty string because we already have an ingress-controller that does the load balancing.
closes #669 

<br/><br/>

<br/><br/>

Run `make install' and edge should deploy successfully.

### DIFF
--- a/edge/templates/edge-service.yaml
+++ b/edge/templates/edge-service.yaml
@@ -14,4 +14,4 @@ spec:
   selector:
     {{ .Values.global.control.cluster_label }}: edge
   sessionAffinity: None
-  type: LoadBalancer
+  type:
--- a/edge/templates/edge-service.yaml
+++ b/edge/templates/edge-service.yaml
@@ -14,4 +14,4 @@ spec:
   selector:
     {{ .Values.global.control.cluster_label }}: edge
   sessionAffinity: None
-  type:
+  type: clusterIP
--- a/voyager/Makefile
+++ b/voyager/Makefile
@@ -6,7 +6,7 @@ voyager:
 	ls
 	helm repo add appscode https://charts.appscode.com/stable/
 	helm repo update
-	helm install voyager-operator appscode/voyager --namespace kube-system --set cloudProvider=minikube --set enableAnalytics=false --set apiserver.enableAdmissionWebhook=false --version v12.0.0-rc.1
+	helm install voyager-operator appscode/voyager --namespace kube-system --set cloudProvider=minikube --set enableAnalytics=false --set apiserver.enableAdmissionWebhook=false --version v12.0.0-rc.1 -f global.yaml
 
 .PHONY: remove-voyager
 remove-voyager:


dgoldstein1:
@rahug20 is this tied to an existing issue?

edge/templates/edge-service.yaml
```diff
@@ -14,4 +14,4 @@ spec:
   selector:
     {{ .Values.global.control.cluster_label }}: edge
   sessionAffinity: None
-  type: LoadBalancer
\ No newline at end of file
+  type:
```
dgoldstein1, line 17: >  This should be removed in favor of a type ClusterIP (or left blank which defaults to clusterIP

I think we should set this as `clusterIP` to be more declarative

rahug20: done

rahug20:
Updated the voyager Makefile as well

chrisbsmith:
@rahug20 @dgoldstein1 where was this tested?  `clusterIP` is not a valid term for a service.  Please submit a PR fixing this ASAP.

```
NOTE: It may take a few minutes for the installation to become stable.
  You can watch the status of the pods by running 'kubectl get pods -w -n default'
sleep 20
(cd edge && make edge)
if [ "edge" == "edge" ]; then \
		echo "run in makefile directory"; \
		helm install edge . -f ../global.yaml; \
	else \
		echo "run in helm-charts top directory"; \
		helm install edge edge -f ../global.yaml; \
	fi
run in makefile directory
Error: Service "edge" is invalid: spec.type: Unsupported value: "clusterIP": supported values: "ClusterIP", "ExternalName", "LoadBalancer", "NodePort"
make[1]: *** [edge] Error 1
make: *** [install] Error 2
```

chrisbsmith:
@dgoldstein1 @rahug20 was this tested in k3d by chance?  I've been using k3d for my Fracking Friday project and using the `ClusterIP` type for edge doesn't seem to work.  We should look into this more as I'm sure there's another alternative.  Right now I'm just curious

PR #715, opened 2020-07-30
K3d v3

EddyIV:
* Update command syntax to k3d v3.0.0
* Removed automatic installation
* Various code cleanup & optimization

### DIFF
--- a/ci/scripts/k3d.sh
+++ b/ci/scripts/k3d.sh
@@ -9,6 +9,18 @@ then
     exit
 fi
 
+output=$(k3d --version)
+len=${#output}
+version=${output:13:1}
+default="3"
+
+if ((version < default)); 
+    then
+        echo '***Please update k3d to v3.0.0 or greater***'
+        exit
+fi
+ 
+
 k3d cluster create $NAME -a 4 -p 30000:10808@loadbalancer && sleep 10
 
 export KUBECONFIG=$HOME/.kube/config
--- a/ci/scripts/k3d.sh
+++ b/ci/scripts/k3d.sh
@@ -9,11 +9,11 @@ then
     exit
 fi
 
-k3d cluster create $NAME -a 4 -p 30000:10808@loadbalancer --update-default-kubeconfig && sleep 10
+k3d cluster create $NAME -a 4 -p 30000:10808@loadbalancer && sleep 10
 
-export KUBECONFIG=$HOME/.k3d/kubeconfig-$NAME.yaml
+export KUBECONFIG=$HOME/.kube/config
 
 echo "Cluster is connected"
 
 echo -e "\nSet KUBECONFIG in your shell by running:"
-echo -e "export KUBECONFIG=$HOME/.k3d/kubeconfig-$NAME.yaml"
+echo -e "export KUBECONFIG=$HOME/.kube/config"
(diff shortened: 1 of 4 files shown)


docs/Deploy with CodeReady.md
```diff
@@ -129,7 +129,7 @@ $ helm dep up data
 $ helm dep up fabric
 $ helm dep up sense
 
-$ helm install secrets secrets -f credentials.yaml
```
chrisbsmith, line 132: Did you mean to remove `credentials.yaml` from this command?  If so, you should also remove the `-f`.  However, I think it's needed.

EddyIV: I honestly don't recall modifying that file at all

docs/Deploy with CodeReady.md
```diff
@@ -114,7 +114,7 @@ Create credential file:
 $ make credentials
 ```
 
-Note: The default Grey Matter install pulls images from the `docker.production.deciphernow.com` repository. If you want to pull images from `docker-dev.production.deciphernow.com` instead, you will have to update all values.yaml files and modify credentials.yaml to use that repository.
+Note: The default Grey Matter install pulls images from the `docker.production.deciphernow.com` repository. If you want to pull images from `docker-dev.production.deciphernow.com` instead, you will have to update all values.yaml files and modify  to use that repository.
```
chrisbsmith, line 117: @EddyIV Restore this file tool?  Maybe there was a find-replace of `credentials.yaml`

chrisbsmith:
@sofiashalotenko do you mind giving this PR a test please?

ci/scripts/k3d.sh
```diff
@@ -1,9 +1,19 @@
...
-curl -s https://raw.githubusercontent.com/rancher/k3d/master/install.sh | TAG=v1.7.0 bash
+# install k3d 3.0.0
+NAME=greymatter
+
+if ! command -v k3d &> /dev/null
+then
+    echo "*** k3d must be installed to start a kubernetes cluster with k3d. Install here: https://k3d.io/#installation ***"
+    exit
+fi
+
```
sofiashalotenko, line 11: Since we're no longer installing the correct version of k3d, it might help if we also have some logic in here to check that the version isn't less than 3.0.

EddyIV: Added

docs/Deploy with K3d.md
```diff
@@ -8,7 +8,7 @@
 ## Local Usage
 
 1. `make k3d` - start a kubernetes cluster locally.
-2. `export KUBECONFIG="$(k3d get-kubeconfig --name='greymatter')"` - configures `kubectl` to use the local k3d cluster.
+2. `export KUBECONFIG=$HOME/.kube/config` - configures `kubectl` to use the local k3d cluster.
```
sofiashalotenko, line 11: Why not change this to `export KUBECONFIG=$(k3d kubeconfig get greymatter)` so it's not pointing to a specific path?

EddyIV: I tried that, but `make credentials` and `make secrets` then return `W0804 15:49:26.443166   22496 loader.go:223] Config not found: ---`

PR #718, opened 2020-08-03
Update v1.2.1

EddyIV:
Updated the following to Greymatter v1.2.1 server versions:
* gm-proxy
* gm-jwt-security
* gm-cli
* gm-dashboard
* gm-slo

***This branch currently has the changes from the k3d-v3 branch (which has not been merged), as I'm using k3d 3.0.0.***


docs/Deploy with K3d.md
```diff
@@ -8,7 +8,7 @@
 ## Local Usage
 
 1. `make k3d` - start a kubernetes cluster locally.
-2. `export KUBECONFIG="$(k3d get-kubeconfig --name='greymatter')"` - configures `kubectl` to use the local k3d cluster.
```
whenhellfreezes, line 11: The more general version here is better because I think you are not gaurenteed that the config will be at this path.

justincely:
@chrisbsmith @zoemccormick is this OBE?

chrisbsmith:
@justincely My guess is "yes", but we need to put eyes on it to confirm

zoemccormick:
@justincely @chrisbsmith yes I think this can be closed

PR #720, opened 2020-08-04
Template spire ca

### LINKED ISSUE #661: template custom SPIRE upstream CA
Right now, the actual chart files need to be edited in order to configure a custom upstream CA for spire.  We should template this to make it easier and not require a local install.

The necessary steps for this configuration are [here](https://docs.greymatter.io/usage/fabric/security/spire/configuration#configure-an-upstream-ca).

There are a couple of options for this:

In the most ideal case, we include a script that would generate all of the necessary certs using [acert](https://github.com/deciphernow/acert) and places them in the correct place in the charts on the fly.  Josh has instructions [here](https://github.com/joshua-rutherford/greymatter-tutorials/tree/master/kubernetes#deploying-a-spire) on the steps for this.

Another (easier) way to do this for now would be to template [this file](https://github.com/DecipherNow/helm-charts/blob/release-2.2/spire/server/templates/server-ca-secret.yaml) and [this file](https://github.com/DecipherNow/helm-charts/blob/release-2.2/spire/server/templates/server-tls-secret.yaml) the same way that we do the secrets chart, by adding the certs to the values file.  It should be noted that we will also need a value for [this configuration](https://github.com/DecipherNow/helm-charts/blob/15227e846565ee656f03d8fd54c27b63289b6140/spire/server/templates/validatingwebhookconfiguration.yaml#L17).

zoemccormick:

closes #661 

This PR adds make command `make spire-custom-ca` and script to automatically generate an upstream CA using acert and replace the necessary secrets files with these new values.

It also allows you to change the spire namespace by setting global.spire.namespace.

<br/><br/>

None 

<br/><br/>

To test:

Have acert installed https://github.com/deciphernow/acert

1. Run `make spire-custom-ca` and follow the prompts - enter the intended spire namespace, and then the args for acert authorities create.

    Update your global.spire.trust_domain and global.spire.namespace (if you changed it) as desired.

2. Run `make install`.

    Everything should come up and the mesh config secrets should use your configured trust domain.  Verify that the mesh 
    comes up correctly and you can access service endpoints through edge.

Exec into the spire server pod `kubectl exec -it server-0 -n spire -c server -- /bin/sh` and run ` /opt/spire/bin/spire-server entry show -registrationUDSPath /run/spire/socket/registration.sock`. Your entries should also have the new trust domain.

To see the certs being presented, pick a service and get its ip `kubectl describe pod {pod-id} | grep IP` and then exec into the data internal container and use openssl to look at the cert: `kubectl exec -it data-internal-0 -c data-internal -- /bin/sh` and run `openssh s_client --connect {IP}:10808`.


ci/scripts/spire-ca.sh
```diff
@@ -0,0 +1,79 @@
...
+
+cd $(dirname "${BASH_SOURCE[0]}")
+read -p "Do you wish to configure a custom upstream CA for SPIRE? [yn] " -n 1 yn
+case $yn in
+    [Yy]* ) echo -e "\nGenerating custom CA for SPIRE"; cacfg $flags;;
+    [Nn]* ) echo -e "\nUsing Quickstart CA for SPIRE"; exit;;
+    * ) echo -e "\nPlease answer yes or no. Defaulting to Quickstart CA"; exit;;
+esac
+
+cmd='acert authorities create '${flags}''
```
chrisbsmith, line 50: This assumes that acert is installed on the machine.

zoemccormick: it does, does that seem like it would be a problem? I could have the script check that acert is installed first

chrisbsmith: Yes, we should not make that assumption.  You can add this command and give the users instructions on how to install acert.   https://github.com/DecipherNow/helm-charts/blob/e97a6209d89d08d4291f856ad96eff4f0f52c150/ci/scripts/k3d.sh#L6-L10

ci/scripts/spire-ca.sh
```diff
@@ -0,0 +1,75 @@
...
+    echo Enter the SPIRE release namespace, or press enter to use the default = \"spire\":
+    read NS
+    if [[ -z "$NS" ]]; then NS=spire; fi
+    echo
+
+    echo Enter the following to configure the authority, or press enter to use the default
+    echo Common Name \(default Acert\):
+    read N; if [[ -z "$N" ]]; then N=Acert; fi
+    echo Organization \(default Decipher Technology Studios\):
+    read O; if [[ -z "$O" ]]; then O=Decipher Technology Studios; fi
```
sofiashalotenko, line 13: At this point in the script I get `./../ci/scripts/spire-ca.sh: line 13: Technology: command not found`.  I think you maybe need to put Decipher Technology Studios in quotes or something?  It seems to be taking Technology as a command.

zoemccormick: @sofiashalotenko strange I don't see this at all - could it be a bash versioning thing? I tried with both `5.0.18(1)` and  `3.2.57(1)`

sofiashalotenko: Hm no I just tried from zsh to bash 3.2.57(1).  I guess if it's just me then we can ignore this.

ci/scripts/spire-ca.sh
```diff
@@ -0,0 +1,75 @@
...
+  --from-literal=registrar.${NS}.svc.key="$(acert leaves export $REGISTRAR_FINGERPRINT -t key -f pem)" > \
+  ../../spire/server/templates/server-tls-secret.yaml
+
+CABUNDLE="$(acert leaves export $REGISTRAR_FINGERPRINT -t authority -f pem | base64)"
+
+sed -i "" "s/caBundle: .*/caBundle: $CABUNDLE/" ../../spire/server/templates/validatingwebhookconfiguration.yaml
+
+acert authorities delete ${AUTH_FINGERPRINT}
+
+echo "SPIRE CA Updated, update global.spire.trust_domain and global.spire.namespace before installing"
```
sofiashalotenko, line 75: What exactly is the trust domain supposed to be changed to?  I left it as quickstart.greymatter.io and it didn't break anything.

zoemccormick: it wont break anything if you leave it - its just now it can be changed to match your new cn/org from the upstream CA - before there were some hard coded instances of quickstart.greymatter.io preventing this

sofiashalotenko: Ah I see.  Then maybe this line could be changed to be a little more specific?  i.e. "optionally update global.spire.trust_domain and update global.spire.namespace if you changed it from the default" (it's a little wordy but maybe a bit more clear)

ci/scripts/spire-ca.sh
zoemccormick: I realized all of my previous test examples change the Org so i was able to recreate it - fixed in my latest commit

PR #723, opened 2020-08-06
Impersonation

### LINKED ISSUE #648: Charts no longer add impersonation filter
In the release-2.1 branch, the proxy would automatically add the impersonation filter to all proxies if the mesh was configured for tls. This appears to have been lost in the updates for release-2.2

This is important because without impersonation, any service could impersonation a user across the mesh with simple header manipulation. @danielpcox identified this with the release-2.1 branch and so a specific edge certificate was created to ensure impersonation across the mesh.

https://github.com/DecipherNow/helm-charts/blob/56c20c835475ad0eafb3cd7fac5b02019f05ecd1/gm-control-api/json/services/proxy.json#L30-L34

### LINKED ISSUE #699: Change JWT filters to use the external facing jwt-security-service filter
The JWT filters currently use the `internal` JWT service rather than the main one.  This was only intended for the internal gm-data, not for all service Auth.   This also means calls to the service don't show up in the dashboard (no card). 

Move all filter chain calls to the main JWT that shows up on the dashboard.

zoemccormick:

<br/><br/>

closes #648 
closes #699 

Adds impersonation filter back for no spire install and switches the default jwt service used for the jwt filter to be the external one

### DIFF
--- a/fabric/control-api/json/services/listener.json
+++ b/fabric/control-api/json/services/listener.json
@@ -83,9 +83,6 @@
           }
       }
   },
-  "gm_impersonation": {
-    "servers": "{{ .Values.services.edge.egressDn }}{{- if not (eq .service.serviceName "dashboard") -}}|{{  .Values.global.sidecar.dn }}{{- end -}}"
-  },
   {{- else }}
     "gm_impersonation": {
       "servers": "{{ .Values.services.edge.egressDn }}{{- if not (eq .service.serviceName "dashboard") -}}|{{  .Values.global.sidecar.dn }}{{- end -}}"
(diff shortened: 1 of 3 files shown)


fabric/control-api/json/services/listener.json
```diff
@@ -82,6 +83,13 @@
           }
       }
   },
+  "gm_impersonation": {
+    "servers": "{{ .Values.services.edge.egressDn }}{{- if not (eq .service.serviceName "dashboard") -}}|{{  .Values.global.sidecar.dn }}{{- end -}}"
+  },
+  {{- else }}
```
sofiashalotenko, line 89: Is this else necessary?  The gm_impersonation section looks the same regardless so I think this can just be a simple 'if' that doesn't include the gm_impersonation section.

zoemccormick: good call that was a test line i accidentally left in!

PR #728, opened 2020-08-20
remove pod name from mongodb SVC

justincely:
3 changes here:
- the mongo svc change.  This allows data to resolve mongo in both GKE and EKS (tested both)
- added a template value for the edge svc type.  This allows dev to still auto provision an ingress point.  I'll update the docs accordingly if this gets merged.
    - to test this, install the edge chart with a loadbalancer: `helm install edge edge --set=global.environment=eks --set=edge.ingress.type=LoadBalancer -f global.yaml` and it'll function as the docs say they do [here](https://docs.greymatter.io/guides/installation-kubernetes#5-accessing-the-dashboard)
- minor, seemingly irrelevant modification to the catalog svc to make it match control-api.  For some reason it didn't work as it was in GKE, now it is fine.

### DIFF
--- a/edge/values.yaml
+++ b/edge/values.yaml
@@ -81,6 +81,9 @@ edge:
       mount_point: /etc/proxy/tls/sidecar/
   # If set, enables ingress TLS on the edge proxy using the secret specified in secret_name
   ingress:
+    # type is used to set how the edge k8s svc resource is created.  For depoyments using NGINX or Voyager or some other
+    # provisioned ingress, leave this as "ClusterIP".   If you want an ingress point automatically created by the cloud provider
+    # set this to "LoadBalancer"
     type: ClusterIP
     secret:
       enabled: true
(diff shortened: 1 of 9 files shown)


edge/values.yaml
```diff
@@ -81,6 +81,7 @@ edge:
       mount_point: /etc/proxy/tls/sidecar/
   # If set, enables ingress TLS on the edge proxy using the secret specified in secret_name
   ingress:
+    type: ClusterIP
```
chrisbsmith, line 84: This is fine, but you should add some documentation to what it does and that LoadBalancer is an available option; with a warning that it will deploy a second LB

justincely: like in `docs/ingress.md`?  or you saying inline to that value?

chrisbsmith: I would say both

chrisbsmith:
@zoemccormick I think the changes in this PR address your comments from #762 .  Agree?  We'll need to have @justincely submit a PR to `release-2.3` as well.

chrisbsmith (approved):
LGTM.  @justincely please submit a PR to `release-2.3` when you can. Though preferably after  PR #761

Issue #646, opened 2020-05-26
GM Data does not need to be included in the default Grey Matter installation


chrisbsmith:
For a standard Grey Matter installation, Grey Matter Data (not the internal Data), does not need to be included.  PR #645 added the option to include it or not, but for completeness, it should be removed all together and transitioned to it's own chart for easy installations for customer specific namespaces.

The issue is that when it is not enabled, the configMap generation still builds the Grey Matter mesh and Catalog API objects and applies them. So when a dashboard is presented, Grey Matter Data still displays, but is shown as down.

zoemccormick:
closed with #664

PR #735, opened 2020-09-02
Add Redis

### LINKED ISSUE #686: update helm-charts to use redis persisters for gm-catalog and gm-control-api
additionally, they should use the redis envoy filter when possible.

bcmendoza:
Closes #686 

Changes:
- Configures Control API and Catalog to use Redis rather than filepersister/Data
- Adds an option for the user to configure an external Redis service (e.g. AWS Elasticache)
- Adds an optional stateful set of a Redis server (if the external Redis is not configured) for persisting configs from Control API and Catalog
- Removes Catalog's use of an internal Data as its config source
- Removes internal JWT, internal Redis, internal Data, and internal Mongo

Note the following work is not covered in this PR:
- The Redis stateful set is not configured to use certs or a password
- The Redis server is not deployed with a sidecar that speaks to it through a network filter

To run:
- Update fabric and sense helm charts
- Launch the mesh in your preferred environment as usual


chrisbsmith (commented):
I am only reviewing this on my phone, but we need to add some documentation to the PR. Specifically, if we are going to advertise that users can leverage AWS Elasticache, we need to document very specifically the minimum IAM role that is needed for this to work. We may even want to write a terraform script to do this for the users.

bcmendoza:
> I am only reviewing this on my phone, but we need to add some documentation to the PR. Specifically, if we are going to advertise that users can leverage AWS Elasticache, we need to document very specifically the minimum IAM role that is needed for this to work. We may even want to write a terraform script to do this for the users.

Sounds good. It's still a WIP, but I will make sure to document the steps for setting up an example connection to AWS Elasticache.

tilleryd:
@bcmendoza Is this still a WIP? We would like to deploy without gm-data in the 1.3 test environment.

fabric/control-api/json/services/listener.json
```diff
@@ -20,10 +20,8 @@
         "X25519:P-256:P-521:P-384"
     ],
     "subject_names": [
-      "spiffe://{{ .Values.global.spire.trust_domain}}/edge"{{- if and (eq .service.serviceName "jwt-security") (.Values.global.data.external.enabled) }},
-      "spiffe://{{ .Values.global.spire.trust_domain}}/data"{{- else if eq .service.serviceName "data-internal"}},
-      "spiffe://{{ .Values.global.spire.trust_domain}}/catalog"{{- else if and (eq .service.serviceName "internal-jwt-security") (.Values.global.data.internal.enabled) }},
-      "spiffe://{{ .Values.global.spire.trust_domain}}/data-internal"{{- end }}
+      "spiffe://{{ .Values.global.spire.trust_domain}}/edge",
+      "spiffe://{{ .Values.global.spire.trust_domain}}/data"
```
zoemccormick, line 24: will data need to be able to route to all services?

fabric/control-api/values.yaml
```diff
@@ -98,10 +98,37 @@ controlApi:
       value: '{{ .Values.global.zone }}'
     gm_control_api_persister_type:
       type: value
-      value: 'file'
-    gm_control_api_persister_path:
+      value: 'redis'
+    gm_control_api_redis_host:
       type: value
-      value: '{{ .Values.controlApi.pvc_mount_point }}'
```
zoemccormick, line 104: can we remove the pvc by default or use a flag to include/not include it since this will now use redis? https://github.com/DecipherNow/helm-charts/blob/b67a59bbe9365774c76b99e9960ad289bf39e4cc/fabric/control-api/templates/control-api.yaml#L150-L156

global.yaml
```diff
@@ -18,13 +18,25 @@ global:
...
+    use_tls: false
+    ca_cert_path: ''
+    server_cert_path: ''
+    server_key_path: ''
+    pass: ''
+    control_api_db: 0
+    catalog_db: 1
+    max_retries: 50
+    retry_delay: '5s'
+
```
zoemccormick, line 39: can some of these be moved out of global values and set as defaults in the redis values? to avoid complicating the globals again

zoemccormick (changes requested):
a couple of questions but everything is functioning as described! awesome to get ride of the catalog-> int data -> int jwt complexities and mesh configs

fabric/control-api/json/services/listener.json
```diff
@@ -20,10 +20,7 @@
         "X25519:P-256:P-521:P-384"
     ],
     "subject_names": [
-      "spiffe://{{ .Values.global.spire.trust_domain}}/edge"{{- if and (eq .service.serviceName "jwt-security") (.Values.global.data.external.enabled) }},
-      "spiffe://{{ .Values.global.spire.trust_domain}}/data"{{- else if eq .service.serviceName "data-internal"}},
-      "spiffe://{{ .Values.global.spire.trust_domain}}/catalog"{{- else if and (eq .service.serviceName "internal-jwt-security") (.Values.global.data.internal.enabled) }},
-      "spiffe://{{ .Values.global.spire.trust_domain}}/data-internal"{{- end }}
+      "spiffe://{{ .Values.global.spire.trust_domain}}/edge"
```
zoemccormick, line 23: this will actually still need a check for jwt so data can connect to jwt

```suggestion
      "spiffe://{{ .Values.global.spire.trust_domain}}/edge"{{- if and (eq .service.serviceName "jwt-security") (.Values.global.data.external.enabled) }},
      "spiffe://{{ .Values.global.spire.trust_domain}}/data"
```

zoemccormick: Im not sure how to handle the case that "external" data is not installed? without the data.external.enabled global option

bcmendoza: `data` is installed by default now... should I have kept the option to disable it?

zoemccormick: im not sure - the ability to disable it was an [issue](https://github.com/DecipherNow/helm-charts/issues/646),  [PR](https://github.com/DecipherNow/helm-charts/pull/664). @chrisbsmith @moosecanswim thoughts on whether or not Data should always be installed by default?

chrisbsmith: No, data should not be installed by default.  Now that we're using Redis, no gm-datas should be installed with a default config.

bcmendoza: Ah okay thanks. I will restore the option to toggle it and add a check for it in this template

zoemccormick (approved):
lgtm!

chrisbsmith:
@bcmendoza when the documentation for using AWS Elasticache Redis is available, can you please link it here?

PR #746, opened 2020-09-24
initial commit to refactor prometheus confs

### LINKED ISSUE #743: Invalid Mesh Configurations generated for Prometheus when using quickstart certs
When running the helm charts without SPIRE enabled, and using our quickstart certs, requests to Prometheus will fail with a `403`.  The prometheus sidecar will log this entry

```
Impersonation Invalid -> USER_DN: CN=quickstart,OU=Engineering,O=Decipher Technology Studios,L=Alexandria,ST=Virginia,C=US | EXTERNAL_SYS_DN:  | SSL_CLIENT_S_DN: CN=quickstart,OU=Engineering,O=Decipher Technology Studios,L=Alexandria,ST=Virginia,C=US
```

Investigation shows that the Prometheus configurations are set to communicate without TLS, from the edge all the way to prometheus. However, the prometheus listener is configured with an active `gm_impersonation` filter. The `403` error is returned when the proxy attempts to authorize the impersonation DN but there is no certificate passed along from the edge.

Two possible solutions:

1. Update the mesh configuration to perform TLS communication from the edge to the prometheus domain; then non-TLS from the cluster to prometheus (since prometheus is not configured for TLS). 

2. Add an additional check of `services.prometheus.secret.enabled` to the `gm.impersonation` filter and configuration

https://github.com/DecipherNow/helm-charts/blob/b67a59bbe9365774c76b99e9960ad289bf39e4cc/fabric/control-api/json/services/listener.json#L30-L34

https://github.com/DecipherNow/helm-charts/blob/b67a59bbe9365774c76b99e9960ad289bf39e4cc/fabric/control-api/json/services/listener.json#L86-L89

Option 1 is the more correct option to take; to force SSL traffic through out mesh. However, it will be a update to how we define the prometheus configurations. In order to not add anymore complexity to our configurations, the prometheus configurations will probably need to move out the `services` loop and into the `special` directory. This will allow us to "hard code" the configurations necessary to enable TLS mesh communications and non-TLS proxy -> service communication.

rahug20:

MR migrated configuration object creation for prometheus under special

Closes #743 

N/A

Configuration is the same, just the steps have been updated 

Create cluster and all the service should be deployed successfully.

### DIFF
--- a/fabric/control-api/config/bootstrap.sh
+++ b/fabric/control-api/config/bootstrap.sh
@@ -66,6 +66,18 @@ for d in */; do
     cd $MESH_CONFIG_DIR/services
 done
 
+# Create configurations objects for Prometheus
+cd $MESH_CONFIG_DIR/special
+echo "Creating special configuration objects for Prometheus"
+create_or_update "domain" domain-prometheus.json
+create_or_update "domain" domain-egress-prometheus.json
+create_or_update "listener" listener-prometheus.json
+create_or_update "listener" listener-egress-prometheus.json
+create_or_update "proxy" proxy-prometheus.json
+create_or_update "cluster" cluster-prometheus.json
+create_or_update "shared_rules" shared_rules-prometheus.json
+create_or_update "route" route-prometheus.json
+
 # The edge service is created last as it links to the clusters of every other service.
 # The edge domain must be created before it can be referenced
 cd $MESH_CONFIG_DIR/special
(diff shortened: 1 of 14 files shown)


chrisbsmith:
@rahug20 you will need to forward port this to the `release-2.3` branch

moosecanswim (approved):
lgtm

Issue #733, opened 2020-08-24
Helm Testing


chrisbsmith:
My fracking friday project (https://github.com/DecipherNow/fracking/blob/master/helm-testing/proposal.md) demonstrated that we can leverage CircleCI to automate the testing of our helm charts. It was a rudimentary test, but one that proves we can test.

This issues assigned to this epic will drive the development of more robust testing of the helm charts.  The current integration test can be found here: https://github.com/DecipherNow/helm-charts/blob/helm-testing/test/greymatter_integration_test.go

This is not an exhaustive list of issues that we can test, just the initial ideas I have.  We should add new issues to this epic as we think of them.

Issue #732, opened 2020-08-24
helm testing: verify that chart versions have been incremented


chrisbsmith:
When we merge a PR for the helm charts, we really should be incrementing the helm chart versions.  We should be able to add a simple test to verify that the version has changed.

Issue #731, opened 2020-08-24
helm testing: verify that the expected services are returned from Catalog


chrisbsmith:
The current test verifies that a query  of the catalog service returns data. However, we know what services should be populated in catalog at the end of the helm installation.  This test should verify that each of the services we expect are listed by Catalog.

Issue #730, opened 2020-08-24
helm testing: verify the correct pods have launched with each helm chart


chrisbsmith:
The integration test deploys each Grey Matter Chart (spire, fabric, edge, data, sense) individually. Currently, it prints out the pods after each chart deploys. We should add verification checks to ensure that each pod we expect to be running is in fact, running.

For example, we know that the fabric chart deploys control, control-api, and jwt. The test should verify all three pods are running.

PR #761, opened 2020-10-16
Add automated testing and linting

chrisbsmith:
## What?

This PR adds automated testing and linting to ever PR that's submitted against a `release-*` branch. It leverages Helm's [chart testing](https://github.com/helm/chart-testing) tool to lint the charts. It lints all charts (individual services and the combo charts [fabric, sense, etc]). It also uses Gruntwork's [terratest](https://github.com/gruntwork-io/terratest/) to automate a deployment of Grey Matter into k3d using helm. (I had to fork their repo and add an update for `helm dep`)

## Why?

Why not?

## How?

In order to achieve this, I had to make a few modifications to the charts:

* all pods now have a `greymatter: <chart>` label. This is used by the helm tests to count the number of pods expected with each chart
* All charts were updated to pass the linter. Some big notes:
  * The Organization config in `Chart.yaml` is compared against GitHub. I changed this to our `greymatter-io` org in anticipation of moving the helm-charts there
  * The website config was removed
  * The `appVersion` has to be a three digit semver, so a `.0` was added to all that were to short
  * **The linter will fail if the chart version is not updated**

The linter and the tests should run whenever a PR is submitted to a `release-*` branch, and the packaging will only occur if the linter and testing pass.  

Definitely would love eyes on this @zoemccormick @tilleryd @dontlaugh @moosecanswim , though technically there is no code change.


.circleci/old.config.yml
```diff
@@ -0,0 +1,261 @@
+# These are the expected environment variables
```
dontlaugh, line 1: We don't need to check this in, imo

chrisbsmith: ha, good catch.

chrisbsmith:
@dontlaugh I meant to add the slack orb to this.  I'll look into this weekend, otherwise I'll put another PR up later if this one makes it in first.  If you know how to do it, I'll take any help

test/greymatter_integration_test.go
```diff
@@ -0,0 +1,277 @@
...
+	"io/ioutil"
+	"log"
+	"strings"
+	"testing"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+
+	"github.com/chrisbsmith/terratest/modules/helm"
```
dontlaugh, line 16: is this a fork?

dontlaugh (commented):
looks good, some questions and comments

dontlaugh:
I've never used an orb, fwiw :shrug:

test/greymatter_integration_test.go
chrisbsmith: Yes, but with my `dependency` module added.

chrisbsmith: https://github.com/chrisbsmith/terratest/blob/master/modules/helm/dependency.go

I need to submit it back to the upstream, but things get busy

dontlaugh: Gotcha, try and get that upstreamed. Otherwise, I'd just wrap the shelling-out right here in this file for `helm dep`.

test/go.mod
```diff
@@ -0,0 +1,11 @@
+module github.com/chrisbsmith/terratest-helm-testing-example/test
```
dontlaugh, line 1: _this_ module path should probably be re-initialized to github.com/DecipherNow/helm-charts

dontlaugh (commented):
LGTM like this but I'd like another set of eyes 

:+1:

chrisbsmith:
@dontlaugh , I believe I addressed all of your concerns.  I've removed the need to use my fork and I've added slack integration so it notifies the #decipher-sres-bots channel.  

Please double check my branch filtering logic in the workflow though.

zoemccormick (approved):
this looks great to me! Everything comes up as expected in the test and in regular behavior.

dontlaugh:
This branch filtering logic looks good to me
```yaml
          branches:
              only: /release-.*/
```

dontlaugh (approved):

test/greymatter_integration_test.go
```diff
@@ -0,0 +1,277 @@
...
+	// Define the path to the catalog service
+	catalogEndpoint := "https://localhost:30000/services/catalog/latest/summary"
+
+	http_helper.HttpGetWithRetryWithCustomValidation(
+		t,
+		catalogEndpoint,
+		&tlsConfig,
+		10,
+		6*time.Second,
+		func(statusCode int, body string) bool {
```
moosecanswim, line 277: You may want to add a statusCode check before trying trying to unmarshal a response.  If the response is bad (maybe a 503) then `body` will be null which will throw an error at `json.Unmarshal`.  if a statusCode check is put in place then we can get emit an error that more accurately reflects whats going on.

test/greymatter_integration_test.go
```diff
@@ -0,0 +1,277 @@
...
+		}
+		log.Fatal(podError)
+	}
+
+	for _, pod := range pods {
+		fmt.Println("Pod: ", pod.Name)
+	}
+}
+
+func verifyCatalog(t *testing.T, kubectlOptions *k8s.KubectlOptions) {
```
moosecanswim, line 243: May want to add a retry similar to how the kubernetes resource checks are done

test/greymatter_integration_test.go
chrisbsmith: So this method `HttpGetWithRetryWithCustomValidation` uses the function to validate the response. It already uses the `statusCode` to catch a non-true value and then retries for 10 times, waiting 6 second between tries.

chrisbsmith: https://github.com/gruntwork-io/terratest/blob/0c509a63364eaee678570c2f64c254c850b2ef19/modules/http-helper/http_helper.go#L78

test/greymatter_integration_test.go
chrisbsmith: The retry logic is actually included in the `HttpGetWithRetryWithCustomValidation` function.

chrisbsmith:
Closes #730 
Closes #731 
Closes #732 
Closes #733

moosecanswim (approved):
LGTM thanks for the clarification on the retry and status checks for the catalog tests!

chrisbsmith:
🤞

PR #763, opened 2020-10-19
close out if statement

zoemccormick:
this is a quick fix - for data not running in eks without this I get 

```bash
Error: parse error at (data/charts/data/templates/mongo.yaml:83): unexpected EOF
helm.go:84: [debug] parse error at (data/charts/data/templates/mongo.yaml:83): unexpected EOF
```

trying to install data


chrisbsmith:
@zoemccormick so you mind reviewing or #761?  It incorporates these changes too

zoemccormick:
oh whoops - yes @chrisbsmith ill close this out and review #761 instead first thing tomorrow

PR #768, opened 2020-10-21
add context to helm test job

chrisbsmith:
One more fix for CircleCI.  The helm jobs were failing because I switched from environment variables to a Context and was not adding  the context to the greymatter-test job.


chrisbsmith:
@moosecanswim @dontlaugh this failed call is due to the nexus being unavailable at the time.

moosecanswim (approved):

PR #772, opened 2020-10-23
k3d version check < 3.1.4

zoemccormick:
set k3d version tag to 3.1.4 following an issue with spire & k3d 3.1.4+. Also adds to the `make k3d` script with this information

see #773


chrisbsmith (approved):
lgmtm

PR #775, opened 2020-10-26
adding resources to the gm services

rahug20:

enabling the resources attributes for gm services

N/A

N/A

deploy the cluster and all the services should be able to deploy successfuly.
When testing in Openshift control should not die down and hit OOM

### DIFF
--- a/fabric/values.yaml
+++ b/fabric/values.yaml
@@ -144,17 +144,16 @@ control:
         type: 'null'
         value: '/app/routes.yaml'
     resources:
-      {}
       # We usually recommend not to specify default resources and to leave this as a conscious
       # choice for the user. This also increases chances charts run on environments with little
       # resources, such as Minikube. If you do want to specify resources, uncomment the following
       # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
-      # limits:
-      #   cpu: 100m
-      #   memory: 128Mi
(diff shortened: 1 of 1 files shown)


fabric/values.yaml
```diff
@@ -144,17 +144,16 @@ control:
...
       # resources, such as Minikube. If you do want to specify resources, uncomment the following
       # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
-      # limits:
-      #   cpu: 100m
-      #   memory: 128Mi
-      # requests:
-      #   cpu: 100m
-      #   memory: 128Mi
+      limits:
+        cpu: 250m
```
chrisbsmith, line 152: Let's give this some room to grow than than hard capping it.
```suggestion
        cpu: 500m
```

fabric/values.yaml
```diff
@@ -144,17 +144,16 @@ control:
...
+        memory: 256Mi
```
chrisbsmith, line 153: Same with memory
```suggestion
        memory: 1Gi
```

chrisbsmith (changes requested):
We shouldn't set the limits to the same values as the requests.  Let's give control some room to grow.

chrisbsmith (approved):
LGTM

PR #780, opened 2020-10-29
Release the Fabric Gov charts

chrisbsmith:
This PR should allow us to package the Fabric Gov chart.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -29,9 +29,27 @@ jobs:
         uses: azure/setup-helm@v1
         id: install-helm
 
-      - name: Run chart-releaser
+      - name: Run Grey Matter Charts
         uses: helm/chart-releaser-action@v1.0.0
         env:
           CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
         with:
           charts_dir: .
+
+      # We need to release the Fabric Gov chart too. So we just update the Fabric chart
+      # and replace the jwt with jwt-gov, then rerun the chart releaser specifically at fabric
+      - run: |
+          cp -f fabric/Chart.yaml ${{ runner.temp }}/Chart.yaml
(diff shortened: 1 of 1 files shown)


chrisbsmith:
@moosecanswim  can I get a review here

.github/workflows/release.yaml
```diff
@@ -29,9 +29,27 @@ jobs:
...
-          charts_dir: .
\ No newline at end of file
+          charts_dir: .
+
+      # We need to release the Fabric Gov chart too. So we just update the Fabric chart
+      # and replace the jwt with jwt-gov, then rerun the chart releaser specifically at fabric
+      - run: |
+          cp -f fabric/Chart.yaml ${{ runner.temp }}/Chart.yaml
+          echo "$(awk 'NR==1,/name: jwt/{sub(/name: jwt/, "name: jwt-gov")} 1' fabric/Chart.yaml)" > fabric/Chart.yaml
+          sed -i.bak-gov -e "s|name: fabric|name: fabric-gov|" -e "s|Grey Matter Fabric|Grey Matter Fabric Gov|" fabric/Chart.yaml
```
moosecanswim, line 44: does this work?  i dont see a place where you are creating a bak-gov file.  `sed -i` does inplace changes on the file so i think this should be `sed -i fabric/Chart.yaml`

moosecanswim (changes requested):

.github/workflows/release.yaml
chrisbsmith: So, per the `sed` [man page](https://linux.die.net/man/1/sed), adding a suffex to `-i` creates a backup on the fly.  So the `sed` command actually creates `fabric/Chart.yaml.bak-gov`.  Though for this we don't really need a backup.

However, I did find that we were not changing the `jwt` chart to use the `jwt-gov` helm charts. So I did push an update that includes that.

moosecanswim (dismissed):

moosecanswim (approved):

Issue #742, opened 2020-09-17
Remove localhost and 0.0.0.0 from all cluster definitions.


joshua-rutherford:
`localhost` resolves to all of the loopback interfaces on the current machine.  in most cases this includes both `127.0.0.1` and `::1`.  as a result, when this is delivered to a sidecar, the sidecar provisions two instances for that cluster, one on each interface. unfortunately, unless the service bound to both explicitly or to all interfaces (i.e., 0.0.0.0 which must be avoided) half of the instances will always be unreachable.  This typically manifests in less than half the requests failing, but the exact percentage is based upon the number of retries and outlier detection.

further, clusters should never use `0.0.0.0` as the host.  i'm not really sure at all what this ultimately what happens with this but it is typically used to denote all interfaces when binding, but clusters are not used to bind so this must be a valid hostname or ip address.

joshua-rutherford:
@danielpcox, @kaitmore, @justincely, @dontlaugh and @zoemccormick, for your awareness given you were all in this conversation.

zakilur:
Cluster definitions have been changed to use `127.0.0.1`.

PR #787, opened 2020-11-04
Removed localhost and 0.0.0.0 from cluster definitions (#742)

zakilur:
Changed the cluster definitions to use `127.0.0.1` instead of `0.0.0.0` or `localhost` as per issue #742 . Confirmed in devinabox that the clusters are using the correct host.


chrisbsmith (approved):
LGTM

PR #790, opened 2020-11-06
Removed waiter-sa error from makefiles (#626)

zakilur:
Issue #626 

Redirected the error stream for the `kubectl get sa waiter-sa` call to `/dev/null`. Also removed some extraneous echo statements.


chrisbsmith (approved):
LGTM

PR #798, opened 2020-11-16
update makefiles to package child charts before they are a make updat…

moosecanswim:
…e-x is applied

add a clean and dep up before running an upgrade

none

none

run make install then change one of the images in sense to `latest` or another version.  then run `make sense-upgrade` from the sense directory


chrisbsmith (approved):
LGTM

PR #824, opened 2020-12-01
822 make sense resources user option

rahug20:

This PR remover hardcoded `resources` section because different environments(OCP3.11, OCP4.6. minishidt) have different capacities and pods do not start because those environments cannot handle the resource capacity

N/A

N/A

Create cluster and it should deploy successfully

### DIFF
--- a/fabric/redis/templates/redis.yaml
+++ b/fabric/redis/templates/redis.yaml
@@ -36,11 +36,12 @@ spec:
         - name: redis
           {{- if and .Values.global.environment (eq .Values.global.environment "openshift") }}
           image: {{ tpl .Values.redis.openshift.image $ | quote }}
+          command: ["redis-server", "--protected-mode no"]
           {{- else if and .Values.global.environment (or (eq .Values.global.environment "kubernetes") (eq .Values.global.environment "eks")) }}
           image: {{ tpl .Values.redis.k8s.image $ | quote }}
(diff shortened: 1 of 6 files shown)


fabric/redis/templates/redis.yaml
```diff
@@ -36,11 +36,12 @@ spec:
         - name: redis
           {{- if and .Values.global.environment (eq .Values.global.environment "openshift") }}
           image: {{ tpl .Values.redis.openshift.image $ | quote }}
+          command: ["redis-server", "--protected-mode no"]
           {{- else if and .Values.global.environment (or (eq .Values.global.environment "kubernetes") (eq .Values.global.environment "eks")) }}
           image: {{ tpl .Values.redis.k8s.image $ | quote }}
+          command: ["redis-server", "--appendonly", "yes"]
```
chrisbsmith, line 42: What is the difference in protected mode vs. appendponly?

chrisbsmith (changes requested):
I don't agree with leaving the resources off here.  Doing so leaves the services up to the defaults of the namespace. Some customers have their namespaces set to super minimal defaults (10mb and 10mi); way too small to run our services.  

Let's find a better alternative to managing these resources.

chrisbsmith:
Closing this PR as stale. If we want to revisit this in the future, we can.

PR #832, opened 2020-12-15
Prometheus role lookup

moosecanswim:

This PR allows the installer to move between role and clusterrole for the prometheus sa without a manual delete step.  It picks up on the difference inside the mesh vs the dashboard values file and will usurp what is currently in the cluster with whatever is in the values file (by removing the old rolebinding then installing a fresh one as part of the upgrade process).

`sense/dashboard/values.yaml` has an additional field prometheus.service_account.kind can be either Role or ClusterRole

<br/><br/>

you can run a `make sense` then change prometheus.service_account.kind and run `make upgrade-sense` and the rolebinding will be deleted before the change.

### DIFF
--- a/sense/values.yaml
+++ b/sense/values.yaml
@@ -301,6 +301,8 @@ dashboard:
     service_account:
       create: true
       name: prometheus-sa
+      # kind specifies either ClusterRole or Role
+      kind: ClusterRole
     resources:
       requests:
         memory: "8Gi"
--- a/sense/Chart.yaml
+++ b/sense/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 1.3.0
 description: A Helm chart to deploy Grey Matter Sense
 name: sense
-version: 3.0.1
+version: 3.0.2
 home: https://greymatter.io
 icon: https://s3.amazonaws.com/decipher-public/grey-matter/branding/grey-matter-mark-rgb.png
 maintainers:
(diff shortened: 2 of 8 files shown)


sense/values.yaml
```diff
@@ -301,6 +301,8 @@ dashboard:
     service_account:
       create: true
       name: prometheus-sa
+      # kind specifies either ClusterRole or Role
+      kind: ClusterRole
```
chrisbsmith, line 305: I know this just adds more stuff to the values file but we should really allow the user to define this.  For instance, in a multi-tenant environment, we probably wouldn't want (or be able) to deploy `ClusterRoles`. Rather they'd be namespace restricted `Roles`. 

Perhaps we could add a flag in `globals.yaml` that defines and drives the scope of the service accounts, roles and rolebindings.

moosecanswim: This pr gives us the ability to bounce between the two.  I agree it should be a role instead of a clusterRole but if we dont have this value then a user will need to manually delete the crb before running an upgrade that includes a role change.

Im fine dropping this PR but we would need to add some documentation to be aware of the case when a customer is performing and upgrade and changing from clusterRole to Role or vice versa

chrisbsmith: I'm not saying we need to drop it, but for now let's open a new issue to track the dynamic update i listed above.

chrisbsmith:
@moosecanswim here's another one that is a bit old at this point. Do we still need it or can we close it?

moosecanswim:
This PR is OBE

I think we can delete it.  We chose to move from a clusterrolebinding to a
rolebinding and this was kinda an upgrade path.

chrisbsmith:
Closing this PR to @moosecanswim's last comment

PR #833, opened 2020-12-16
fix flags for elk stack

zoemccormick:
This adds the last couple of flags to make observables install correctly


dgoldstein1 (approved):
tested and approved

PR #838, opened 2021-01-04
update data values file and increment the chart and parent charts acc…

moosecanswim:
…ordingly

<br/><br/>

<br/><br/>

<br/><br/>

<br/><br/>

### DIFF
--- a/data/Chart.yaml
+++ b/data/Chart.yaml
@@ -16,6 +16,6 @@ dependencies:
   # todo: combine into one chart
   - name: gm-data
     repository: file://./gm-data
-    version: '3.0.4'
+    version: '3.0.5'
     alias: data
     condition: global.data.external.enabled
--- a/data/gm-data/Chart.yaml
+++ b/data/gm-data/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v2
-appVersion: 1.0.0
+appVersion: 1.1.4
 description: A Helm chart to deploy Grey Matter Data
 home: https://greymatter.io
 icon: https://s3.amazonaws.com/decipher-public/grey-matter/branding/grey-matter-mark-rgb.png
@@ -10,4 +10,4 @@ maintainers:
 - email: engineering@greyatter.io
   name: greymatter-io
 name: gm-data
-version: 3.0.4
+version: 3.0.5
--- a/data/gm-data/values.yaml
+++ b/data/gm-data/values.yaml
@@ -37,7 +37,7 @@ global:
 data:
   # Name used for the deployment and service resources
   name: data
-  version: 1.1.3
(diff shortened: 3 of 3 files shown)


chrisbsmith (commented):
@moosecanswim why are we just incrementing the chart files when there's no change to the charts themselves?

moosecanswim:
I incremented the data version which means we should increment the chart.yaml version so the change gets published.  data is wrapped in gm-data parent chart so that has to get updated too.  Tho it does not really matter cause this chart will be pulled out of this repo soon

chrisbsmith:
> I incremented the data version which means we should increment the chart.yaml version so the change gets published. data is wrapped in gm-data parent chart so that has to get updated too. Tho it does not really matter cause this chart will be pulled out of this repo soon

I get the versioning aspect, but was more curious why we were incrementing the version with no code changes.

moosecanswim:
> I get the versioning aspect, but was more curious why we were incrementing the version with no code changes.

I dont think we should treat the values files different from the code.  In an ideal world the people deploying these services would understand a difference but from my experience they dont get that.

if we increment the chart version then we get the ability to say 

hey paul if you want data 1.1.4 run `helm install data decipher/data --version 3.0.5` if you want data 1.1.2 use `--version 3.0.5`

seems easier than saying hey paul install this version then change the value data.image.

chrisbsmith:
> > I get the versioning aspect, but was more curious why we were incrementing the version with no code changes.
> 
> I dont think we should treat the values files different from the code. In an ideal world the people deploying these services would understand a difference but from my experience they dont get that.
> 
> if we increment the chart version then we get the ability to say
> 
> hey paul if you want data 1.1.4 run `helm install data decipher/data --version 3.0.5` if you want data 1.1.2 use `--version 3.0.5`
> 
> seems easier than saying hey paul install this version then change the value data.image.

I agree.  This change then is only to increment the `appVersion` then?  I suppose there was another PR that updated the values files for the `1.1.5` docker images

chrisbsmith:
@moosecanswim can you please look into this again.

moosecanswim:
@chrisbsmith I think we can drop this pr since we're not deploying data with the mesh.  We should probably drop data entirely from the repo

chrisbsmith:
@moosecanswim we didn't get this merged in before the update to `main`. Do we still need it?

moosecanswim:
@chrisbsmith correct we did not merge this in

chrisbsmith:
But do we need it?

moosecanswim:
If we are pulling data out of the core helm charts then we dont need it.

chrisbsmith:
Closing this PR as we'll be removing GM Data from the core GM Helm Charts.

PR #848, opened 2021-01-12
Publish spire server agent

moosecanswim:
Switch the spire release with the greymatter release to force publish.  The release will fail on greymatter but thats ok since those charts are already published.


chrisbsmith (approved):
lgtm

PR #869, opened 2021-01-22
Update the linters to catch when an umbrella chart needs to be updated

### LINKED ISSUE #854: Linting failed to detect chart change
Going back to PR https://github.com/greymatter-io/helm-charts/pull/840 it appears that a change to the Fabric chart was made without the linter detecting that the version needed to be updated. Here is the failing Action https://github.com/greymatter-io/helm-charts/actions/runs/472265935.

chrisbsmith:
This PR fixes the umbrella chart linting.  It adjusts two things:
* installs helm and runs a `helm dep up` for the umbrella charts
* updates the umbrella config file 

You can see a successful capture of the need to update an umbrella chart here: https://github.com/greymatter-io/helm-charts/runs/1750370868?check_suite_focus=true

And then a successful run of the linter, https://github.com/greymatter-io/helm-charts/actions/runs/504268990

closes #854

### DIFF
--- a/.chart-testing/fabric-sense-data.yaml
+++ b/.chart-testing/fabric-sense-data.yaml
@@ -1,6 +1,3 @@
-charts:
-  - fabric
-  - sense
-  - data
-  - spire
+chart-dirs:
+  - .
 target-branch: release-2.3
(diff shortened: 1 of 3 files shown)


moosecanswim (approved):
@chrisbsmith LGTM.  Thanks for providing the link to the helm chart where there is a fabric diff.

PR #878, opened 2021-01-28
Global replicas

moosecanswim:

This allows us to create production ready mesh (meaning multiple replicas).  You can also set the default_replicas.

```
  release:
    production: false
    default_replicas: 3
```
If you set production to false then 3 instances of each service (not control api) will be created

added a note to `### Installing`

I ran make install in a devinabox cluster and this worked.  

set global.release.production to true then run a standard installation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,8 @@ helm install edge greymatter/edge --set=edge.ingress.type=LoadBalancer -f global
 helm install sense greymatter/sense -f global.yaml --set=global.waiter.service_account.create=false
 ```
 
+> If you would like to scale to a production ready mesh set global.release.production to true in `global.yaml`
+
 ### Viewing the Grey Matter Application
 
 At this point, you can verify that Grey Matter was installed successfully by opening your browser and pointing it to `https://localhost:30000` and verify that all six services are running.
--- a/edge/Chart.yaml
+++ b/edge/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.5.1
 description: A Helm chart to deploy Grey Matter Edge
 name: edge
-version: 3.0.2
+version: 3.0.3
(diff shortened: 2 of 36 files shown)


chrisbsmith:
> If you set production to false then 3 instances of each service (not control api) will be created

What happens if i set `production` to true?

moosecanswim:
> > If you set production to false then 3 instances of each service (not control api) will be created
> 
> What happens if i set `production` to true?

It will set the replicas of everything except the api to the dumber of replicas specified in default_replicas

chrisbsmith:
> > > If you set production to false then 3 instances of each service (not control api) will be created
> > 
> > 
> > What happens if i set `production` to true?
> 
> It will set the replicas of everything except the api to the dumber of replicas specified in default_replicas

OK, maybe your initial comment is incorrect then.  Here's what I expect:

- when `production` is false (this should be the default): we deploy a single instance of all pods
- when `production` is true, we deploy `default_replicas` (default of 3) number of replicas

Is this accurate?

moosecanswim:
> > > > If you set production to false then 3 instances of each service (not control api) will be created
> > > 
> > > 
> > > What happens if i set `production` to true?
> > 
> > 
> > It will set the replicas of everything except the api to the dumber of replicas specified in default_replicas
> 
> OK, maybe your initial comment is incorrect then. Here's what I expect:
> 
>     * when `production` is false (this should be the default): we deploy a single instance of all pods
> 
>     * when `production` is true, we deploy `default_replicas` (default of 3) number of replicas
> 
> 
> Is this accurate?

Oh whoops.  Yes.  If you set production:true then it will use number specified by default replicas.  If it is false then you will use what is defined in the individual services values.yaml.

If nothing is set anywhere then it will default to 1

zakilur (approved):
LGTM

PR #888, opened 2021-02-04
Helm release names

moosecanswim:
updates the helm release names that the make files use.

**This has a change to global.yaml**.  a customer field has been added under global.releases.  This is used to create the helm release name as well as an identifier inside a mesh config.

to test install the mesh from the makefiles like normal `make secrets && make install`

do a `helm ls` and you will see all the regular releases in the form `<chart>-<customer>-<random number>`

You should be able to hit the endpoint as you normally do.

To test that the configmap can protect the mesh:

> (if you changed the customer value then replace "development" with that name)

First copy the configmap so we can easily restore it:
 `kc get configmap greymatter-mesh-identity-development -o yaml > mesh-identity.yaml`

Now delete that configmap
`kc delete configmap greymatter-mesh-identity-development`

Run `make uninstall` **This will fail** since the config map is what provides the random sting

Add that configmap back in `kc apply -f mesh-identity.yaml` 

Run `make uninstall`

These commands are also in the readme.

This will help avoid the outage we had last week.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,13 +4,13 @@ SHELL := /bin/bash
 # `make credentials` to build out credentials with user input
 # `make secrets` deploys the credentials
 
+.PHONY: minikube k3d check-secrets install-spire install uninstall secrets remove-secrets credentials observables remove-observables spire-custom-ca lint-subcharts lint-edge-secrets lint-umbrella-charts lint
+
 K3D?=false
 
-.PHONY: minikube
 minikube:
 	./ci/scripts/minikube.sh
 
-.PHONY: k3d
 k3d:
 	./ci/scripts/k3d.sh
   K3D=true
@@ -38,21 +38,18 @@ dev-dep: clean
 	(cd fabric && make package-fabric)
 	(cd sense && make package-sense)
 
-.PHONY: check-secrets
 check-secrets:
 	$(eval SECRET_CHECK=$(shell helm ls | grep secrets | awk '{if ($$1 ~ /secrets*/) print "present"; else print "not-present"}'))
 	if [[ "$(SECRET_CHECK)" != "present" ]]; then \
 		(make secrets); \
 	fi
 
-.PHONY: install-spire
 install-spire:
 	$(eval IS=$(shell cat global.yaml | grep -A3 'spire:'| grep enabled: | awk '{print $$2}'))
 	if [ "$(IS)" = "true" ]; then \
 		(cd spire && make spire); \
 	fi
 
-.PHONY: install
 install: dev-dep check-secrets install-spire
 	(cd fabric && make fabric)
 	sleep 20
@@ -66,7 +63,6 @@ install: dev-dep check-secrets install-spire
 	(make reveal-endpoint)
 
 .IGNORE: uninstall
-.PHONY: uninstall
 uninstall:
 	-(cd spire && make remove-spire)
 	-(cd fabric && make remove-fabric)
@@ -94,51 +90,41 @@ template: dev-dep $(BUILD_NUMBER_FILE)
 	(cd sense && make template-sense && cp $(OUTPUT_PATH)/* ../$(OUTPUT_PATH)/)	
 
 
-.PHONY: secrets
 secrets:
 	cd secrets && make secrets
 
-.PHONY: remove-secrets
 remove-secrets:
 	helm uninstall secrets
 
-.PHONY: credentials
 credentials:
 	cd secrets && make credentials
 
 EKS?=false
 OBSERVABLES_NAMESPACE?=observables
 
-.PHONY: observables
 observables:
 	cd observables && \
 	make check-namespace NAMESPACE=$(OBSERVABLES_NAMESPACE) && \
 	make check-secrets NAMESPACE=$(OBSERVABLES_NAMESPACE) && \
 	make install-observables NAMESPACE=$(OBSERVABLES_NAMESPACE) EKS=$(EKS)
 
-.PHONY: remove-observables
 remove-observables:
 	cd observables && \
 	make destroy-observables NAMESPACE=$(OBSERVABLES_NAMESPACE)
 
-.PHONY: spire-custom-ca
 spire-custom-ca:
 	cd spire && make custom-ca
 
-.PHONY: lint-subcharts
 lint-subcharts:
 	@echo "Lint Fabric, Sense, and Spire subcharts"
 	ct lint --config .chart-testing/services.yaml
 
-.PHONY: lint-edge-secrets
 lint-edge-secrets:
 	@echo "Lint Edge and Secrets"
 	ct lint --config .chart-testing/edge-secrets.yaml 
 
-.PHONY: lint-umbrella-charts
 lint-umbrella-charts:
 	@echo "Lint top level charts"
 	ct lint --target-branch release-2.3 --chart-dirs .
 
-.PHONY: lint
 lint: lint-subcharts lint-edge-secrets lint-umbrella-charts
--- a/Makefile
+++ b/Makefile
@@ -39,8 +39,8 @@ dev-dep: clean
 	(cd sense && make package-sense)
 
 check-secrets:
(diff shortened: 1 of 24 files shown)


chrisbsmith:
@moosecanswim Rather than defining the name as `-dev`, what if we made a random string per deployment?  Something like `cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 10 | head -n 1`.  (Though this doesn't work on a mac.)

moosecanswim:
@chrisbsmith we would need to set an environment value on the users comp so that we could reference it when its time to uninstall

chrisbsmith:
@moosecanswim or grab it from a `helm list`, or set a temp hidden file with the value.  just something to think about to try and randomize things.

moosecanswim:
@tilleryd , @chrisbsmith would like you to review this as well since it does have some user experience integration.

moosecanswim:
The origional issue from clickup 
```
Update the makefiles to change the release name from fabric`, `sense`, `edge`, `secrets`  to the form of fabric-<customer> default to fabric-dev .  This is will ensure that we do not accidentally destroy a production mesh if in a customers cluster and run make uninstall .
```

edge/Makefile
```diff
@@ -1,7 +1,10 @@
 SHELL := /bin/bash
 mkfile_path := $(abspath $(lastword $(MAKEFILE_LIST)))
 current_dir := $(notdir $(patsubst %/,%,$(dir $(mkfile_path))))
-NAME-EDGE := edge
+CUST := $(shell cat ../global.yaml | docker run -i docker.greymatter.io/internal/yq:2.4.1 -r '.global.release.customer' )
```
chrisbsmith, line 4: I'm really on the fence about forcing this hard dependency on another docker image.  Can you explore another option using grep and/or sed or any other tool that could natively be on a machine?

moosecanswim: the hard thing with yq and jq is they are easy to install and light but there are different versions.  The yq that works best is one that translates yaml to json then wraps json... we can use wither sed or awk but yq/jq are just so much cleaner (and will not be screwed up if we make changes to the global.yaml).

I chose docker because we can almost certainly guarantee that it will be on developers computers (and that docker image is hosted by us).

moosecanswim:
@chrisbsmith @tilleryd, Could i get a thumbs up or down?

Makefile
```diff
@@ -125,4 +124,22 @@ remove-observables:
...
+	@echo "Lint Fabric, Sense, and Spire subcharts"
+	ct lint --config .chart-testing/services.yaml
+
+.PHONY: lint-edge-secrets
+lint-edge-secrets:
+	@echo "Lint Edge and Secrets"
+	ct lint --config .chart-testing/edge-secrets.yaml 
+
+.PHONY: lint-umberela-charts
+lint-umberela-charts:
```
tilleryd, line 140:
```suggestion
lint-umbrella-charts:
```

Makefile
```diff
@@ -125,4 +124,22 @@ remove-observables:
...
+	@echo "Lint top level charts"
+	ct lint --target-branch release-2.3 --chart-dirs .
+
+.PHONY: lint
+lint: lint-subcharts lint-edge-secrets lint-umberela-charts
```
tilleryd, line 145:
```suggestion
lint: lint-subcharts lint-edge-secrets lint-umbrella-charts
```

fabric/Makefile
```diff
@@ -1,6 +1,9 @@
 SHELL := /bin/bash
-NAME-FABRIC:=fabric
 CHART-FABRIC := .
+YQCMD := docker run -i docker.greymatter.io/internal/yq:2.4.1
+CUST := $(shell cat ../global.yaml | $(YQCMD) -r '.global.release.customer' )
+RAND := $(shell kubectl get configmap greymatter-mesh-identity-$(CUST) -o jsonpath='{.data.rand_identifier}')
+NAME-FABRIC := fabric-$(CUST)-$(RAND)
```
tilleryd, line 6: Can these instructions be moved to a global function? The signature would take a chart name variable to be used in the customer name variable and value. Then you can replace the copy pasta with this function throughout.

tilleryd (changes requested):
see comments

global.yaml
```diff
@@ -23,6 +23,7 @@ global:
   
   release:
     # release.production: true will increase all (except control-api) replicas to the default_replicas.  Otherwise the default for each chart will be used
```
tilleryd, line 25: I'm going to comment on an unchanged line here. To me the term default here is overloaded and this comment doesn't clearly explain the difference between default_replicas and default.

global.yaml
```diff
@@ -23,6 +23,7 @@ global:
...
+    customer: development
```
tilleryd, line 26: Can we use a better default customer name? `development` doesn't sound like a customer. Perhaps `greymatter`.

tilleryd:
I think the naming of the spire charts is strange.

```
spire-development-ir2msj-agent 
spire-development-ir2msj-server
```

I believe they should read:

```
spire-agent-development-ir2msj 
spire-server-development-ir2msj
```

moosecanswim:
@tilleryd I refactored the makefiles to reuse code in the `helpers.mk` file.  I also moved .PHONY to single lines

moosecanswim:
@chrisbsmith you can set the yq command in the `helpers.mk` file which is imported by each of the subchart makefiles

global.yaml
```diff
@@ -19,10 +19,11 @@ global:
     # The label Control uses to find pods to include in the mesh
     cluster_label: greymatter.io/control
     # Comma delimited string of namespaces for control to monitor. Also used by prometheus.
-    additional_namespaces:
+    additional_namespaces: ""
   
   release:
     # release.production: true will increase all (except control-api) replicas to the default_replicas.  Otherwise the default for each chart will be used
+    customer: greymatter-dev
```
tilleryd, line 26: @moosecanswim I am now thinking this should be an empty string by default. Since we use these charts in https://docs.greymatter.io/installation/installation-kubernetes and "customer" is very specific to operations, I think we should keep the default as is. Thus there will be no changes to our docs and the output will be the same as it has been for quickstart deployments and development.

tilleryd:
@moosecanswim I think a better implementation is to not create the `greymatter-mesh-identity-x` configmap if `production: true`. Currently, one would need to intentionally delete the configmap after install which still leaves the possibility for unintended deletion of the mesh.

global.yaml
tilleryd: @moosecanswim thoughts?

moosecanswim: I disagree with this.  The whole motivation behind this is to add an extra layer to prevent unintentional deletion of the mesh due to a dev using the wrong k8 context.  Lets say you are connected to a customer cluster but think you are connected to k3d.  if you run a `make uninstall` as it stands the makefile will uninstall the default releases `fabric` `sense` `edge` etc.  Since they are uninstalled there is no way to rollback or easily recover the mesh.

 This is what we think happened a couple weeks ago with the di2e mesh.

The way this pr protects against this is by ensuring that the release name is unique per installation/ customer.  This only really affects those using makefiles; customers deploying the mesh using our docs site will not be affected.  However this is laying the foundation for wrapping our helm charts in a go wrapper.  Using a wrapper with similar logic to the makefiles will allow us allow us to give these unintentional deletion protections to the customers.

The work that i did yesterday does automatically delete the configmap upon install.  This means that you will need to run a command to restore the `mesh-identity` configmap before running the uninstall.

tilleryd: For production you need to have the value set as a standard operating procedure in my mind. You would change it anyways since `greymatter-dev` shouldn't be the default.

tilleryd: @chrisbsmith thoughts? I don't want to be too pedantic and hold this up

moosecanswim: If we leave it blank then there would still be a rand portion to the release.

tilleryd: Ok, I'm probably splitting hairs here.

tilleryd:
@moosecanswim you just need to fix linting before I can approve

tilleryd (approved):
@moosecanswim 👍

PR #896, opened 2021-02-11
add gitbook sync workflow to 2.4

moosecanswim:
brings the same gitbook sync from release-2.3 to release-2.4 but points the target branch to `1.4-beta`


chrisbsmith (approved):
lgtm

PR #899, opened 2021-02-17
Mesh redis password

moosecanswim:

This adds a password to the mesh redis instance.  This pr also adds the password for use to catalog and control

I updated the logic so the envars are set from values that mimic the external_redis config. Other than that global.yaml had a secret_name added.

none is really needed since the secret is random unless specified and its purely internal

run `make secrets` and `make install` it will work.  If you want to look at the secret you can run `kc get secret mesh-redis-password -o yaml` you will see it is random.  If you want to set your own password then in secrets/values.yaml edit `mesh_redis`


chrisbsmith:
@moosecanswim , did you also test a helm upgrade?

fabric/redis/values.yaml
```diff
@@ -15,6 +15,8 @@ redis:
   image_pull_policy: IfNotPresent
   # Uses global.image_pull_secret if true
   private_image: false
+  secret:
+    secret_name: mesh-redis-password
```
chrisbsmith, line 19: This would need to match the values from global.yaml

fabric/redis/values.yaml
```diff
@@ -28,3 +30,8 @@ redis:
     requests:
       cpu: 100m
       memory: 128Mi
+  envvars:
+    redis_password:
+      type: secret
+      secret: '{{ .Values.redis.secret.secret_name | default "mesh-redis-password" }}'
```
chrisbsmith, line 36: possibly here too for the default.

fabric/values.yaml
```diff
@@ -157,8 +157,18 @@ control:
 
 
 control-api:
-  redis:
+  mesh_redis:
     name: mesh-redis
+    secret_name: mesh-redis-password
```
chrisbsmith, line 162: same naming convention

global.yaml
```diff
@@ -38,13 +38,14 @@ global:
   # If an external Redis service is not needed, an internal Redis service is launched by setting `disabled` below to true
   external_redis:
     disabled: true
+    secret_name: mesh-redis-password
```
chrisbsmith, line 41: another

chrisbsmith (changes requested):
@moosecanswim I like what you've done here. I have not run it yet so I can't vouch that it works, but I'd like to see us take more ownership of the resources we lay down.  To start that, I'd like to put a `greymatter-` prefix on our resources. Since you're creating a new secret here, this is a great opportunity to do that.

secrets/templates/docker-secret.yaml
```diff
@@ -13,7 +13,6 @@
 {{- end }}
 
 {{- $imagePullSecret := (lookup "v1" "Secret" .Release.Namespace .Values.global.image_pull_secret) }}
```
chrisbsmith, line 15: By removing the line below, it seems this lookup is no longer needed.

chrisbsmith (changes requested):
still haven't tested it personally, but do have a comment on the docker secret.

chrisbsmith (approved):
Thanks for making the updates @moosecanswim .  Based on the testing from the PR commit, this LGTM

chrisbsmith (approved):
lgtm

PR #907, opened 2021-02-26
Helm release names (#888)

moosecanswim:
Adding in some code that was merged into release-2.3


justincely (approved):
looks reasonable to me

PR #917, opened 2021-03-15
remove setting of NodeID to default

justincely:

1. Each sidecar is currently set to announce it's nodeID as "default".  This is from some legacy behavior with `gm-xds` and should no longer be used.  It's not harmful by itself, but simply means in the logs it's impossible to uniquely identify an instance.  Right now all N instances of "gm-catalog" would announce with `nodeID: default` and `cluster: catalog`.  If this is omitted, then a random string is generated for each, so each node gets a unique `nodeID: 234j98sadflk34j`.  This means log messages can be associated to unique instances.  

<br/><br/>

2. What **changes to custom.yaml** are required?  N/a

<br/><br/>

3. Have you documented any additional setup steps or configurations options? n/a

<br/><br/>

4. Deploy with charts.  

deploy

<br/><br/>


moosecanswim (approved):
LGTM.  @justincely could you also make this change to the release-2.4 branch.  And this should be something we let the solutions leads know as well so they can update their application helm-charts so we can node level logging in production.

justincely:
yeah sure thing.  I'll get 2.4 next.

Do you have an avenue for dissemination to the solutions leads?

PR #924, opened 2021-03-22
Changed CN=quickstart references to greymatter.user

zakilur:
**Please provide enough information so that others can review your pull request:**

1. Explain the **details** for making this change. What existing problem does the pull request solve? If it resolves an existing issue, be sure to use a Github [keyword](https://help.github.com/en/articles/closing-issues-using-keywords) to automatically close it.

This change is in response to issue #923 . Moving away from the quickstart certs required changing the CN to match that provided by the `greymatter-user-cert` secret.
<br/><br/>

2. What **changes to custom.yaml** are required?

None
<br/><br/>

3. Have you documented any additional setup steps or configurations options?

None
<br/><br/>

4. Demonstrate the code is solid. Example: The exact commands you ran and their output, screenshots / videos if the pull request changes UI.

Changes have been verified by deploying Fibonacci service to a local K3D cluster.
Confirmed the following as working:

- Fibonacci service, including mesh configs, was successfully deployed.
- Fibonacci card on the dashboard was displaying properly and was showing metrics from Prometheus
- SLOs were tested successfully; they displayed properly in the dashboard and were confirmed in Postgres
- Business Impact was tested successfully; Changes in business impact were reflected in Redis and the catalog entry

<br/><br/>

PR #927, opened 2021-03-23
If global certs enabled, sidecar-certs should match

chrisbsmith:

When `global.global_certs.enabled` is set to true, the `sidecar-certs` should match the values provided for the `global_certs`. This fixes that.

none

none needed

Helm templates verify that the certs are legit.

This PR will need to be rebased after #924 is merged.

### DIFF
--- a/secrets/Chart.yaml
+++ b/secrets/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: A Helm chart to deploy secrets
 name: secrets
-version: 2.4.1
+version: 2.4.2
 home: https://greymatter.io
 icon: https://s3.amazonaws.com/decipher-public/grey-matter/branding/grey-matter-mark-rgb.png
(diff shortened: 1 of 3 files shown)


moosecanswim:
@chrisbsmith instead of creating a second credential when making the global certs it might be cleaner to just set the secret to global certs when setting up the volume https://github.com/greymatter-io/helm-charts/blob/ed09de7cfb6e857fd908ee6ebd8f5e7078a298dd/sense/catalog/templates/_sidecar_volumes.tpl#L4

chrisbsmith:
@moosecanswim , I did think about this, but if you do a search for "sidecar-certs" throughout the charts, you'll find a lot of references to it.  My change (b/c I need it for a customer install), is a simple update that doesn't have many downstream impacts.

moosecanswim (approved):
LGTM

PR #928, opened 2021-03-30
Allow the Edge Name to be templated

chrisbsmith:

The Grey Matter edge chart is one that can be reused in scenarios where we want to add an Edge node to a specific service. As it stands, the edge name is hard coded so it will never work.  We need to allow the edge name to be templated.

n/a

n/a

PR tests.


jimgaspari (approved):
The changes are valid and look good.

PR #950, opened 2021-05-13
Upgrade to GM RC versions

bcmendoza:
[ch-2576]
[ch-2170]

- Upgrades Catalog to 2.0 and replaces the catalog-init process with a configMap file that seeds data into Catalog on startup.
- Tags all release candidates


dontlaugh:
@zoemccormick Should we close this PR, do you think?

zoemccormick:
@dontlaugh I can update the versions once the GA releases are all cut

chrisbsmith:
The SREs are tracking this as PR that's going to drop 1.6 into place.  Is this a correct assumption?

zoemccormick:
yes @chrisbsmith

catalog.sh
```diff
@@ -0,0 +1,286 @@
+#!/usr/bin/env bash
```
chrisbsmith, line 1: I don't think we want to check this file into the main branch.

ci/scripts/gen-certs.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
chrisbsmith, line 1: I'm all for automation, but this script assume that the user has `acert` installed and in the path. If we keep it here, recommend adding a `command` check to see if it exists. Something like this will do

```
if ! command -v acert &> /dev/null
then
    echo "acert could not be found"
    exit
fi
```

edge/templates/_edge_volumes.tpl
```diff
@@ -12,7 +12,12 @@
 {{- define "edge_ingress_volumes" }}
 - name: edge-ingress
   secret:
+    {{- if .Values.global.global_certs.enabled }}
```
chrisbsmith, line 15: We need to review this because we allow for the users to specify a unique ingress cert.

https://github.com/greymatter-io/helm-charts/blob/b276c702444f2889554254201dedcf468776cd31/secrets/values.yaml#L1017

edge/values.yaml
```diff
@@ -13,6 +13,9 @@ global:
   route_url_name: greymatter
   remove_namespace_from_url: 'false'
 
+  global_certs:
```
chrisbsmith, line 16: I'm not 100% certain this is necessary since Edge is a unique case.

@moosecanswim  please review.

fabric/jwt/values.yaml
```diff
@@ -98,7 +98,7 @@ jwt:
     {
       "users": [
           {
-        "label": "CN=quickstart,OU=Engineering,O=Decipher Technology Studios,L=Alexandria,ST=Virginia,C=US",
+        "label": "CN=istio-quickstart (Intermediate),OU=Engineering,O=Decipher Technology Studios,L=Alexandria,ST=Virginia,C=US",
```
chrisbsmith, line 101: Is this really the name of the cert we want to publish?

sense/catalog/templates/catalog.yaml
```diff
@@ -17,16 +17,6 @@ spec:
         greymatter: sense
     spec:
       serviceAccountName: {{ .Values.global.waiter.service_account.name }}
-      initContainers:
```
chrisbsmith, line 20: I'm all about removing the complexity of init containers!

sense/catalog/values.yaml
```diff
@@ -86,6 +86,9 @@ catalog:
       cert: server.crt
       key: server.key
   envvars:
+    debug:
+      type: value
+      value: 'true'
```
chrisbsmith, line 91: Let's make sure we turn this to `false` when we publish. We don't want to have our production services running at `debug` level.

chrisbsmith (changes requested):
Thanks for workings this update @bcmendoza . I do have a few comments that we need to address.

@moosecanswim and @zakilur  please make sure you get a review of this, especially the global cert usage on the edge.

zoemccormick:
@chrisbsmith this stuff was originally just meant to be used on the istio deployment when the quickstart certs expired - we can either try to clean these scripts up or include instructions on how to generate all of the certs and where to put them if we're worried about the global certs

moosecanswim:
@zoemccormick could yall clean up the scripts.  Adding more scripts just makes it harder to maintain customer environments.  If the cert is unique to the installation could it be generated with an init job?

global.yaml
```diff
@@ -64,15 +64,15 @@ global:
...
 
   # Configure services to use the same certs
   global_certs:
-    enabled: false
+    enabled: true
     certificates:
       # if from file is true the there must be a ca.crt, server.crt, and server.key files in the secrets/<$path> directory
       from_file:
-        enabled: false
+        enabled: true
```
moosecanswim, line 75: Using from file will not work for customers that do not have access to the helm charts source code.  At this time most customers do a `helm repo add greymatter <repo url>` then run a `helm install <x> greymatter/<chart>`.  With this as true the files will not exist so there will be an empty secret.

chrisbsmith:
@zoemccormick we we're under the impression this PR was making the move to 1.6. If that is not the case we need to start over, which tbh, may not be a bad thing. I'm not really sure of the state of the release-2.4 branch but I do know there are a ton of PRs that went into release-2.3 and not 2.4. 

We will need to define a better process for this. I'm debating moving everything to a "main" branch and then using tags to version control. I think this will prevent headache in the future and allow us to easily decouple the helm versions from GM version. 

@dontlaugh would love your input here as well. You have a solid understanding of this

kaitmore:
@chrisbsmith I do think that we should start over and I'm going to close this PR now. This branch is outdated and has a lot of our environment specific testing stuff. release-2.4 is also out of date so we should focus on getting that to parity with release-2.3 first. This PR can just serve as a reference for 1.6 changes.

chrisbsmith:
@zakilur FYI

PR #967, opened 2021-07-20
Update testing 1.6

moosecanswim:
This is somewhat of a rebuild and refactor of the helm charts templating.  Errors will be caught earlier instead of running through all tests if they fail.  The test flag `-failfast` and the updated error catching in the tests are used to prematurely finish the testing if a faliure is found.

installation and releases have been abstracted out so we dont need to make changes to each installation test... just one.

## To test locally:
`make k3d`
`make secrets`
`make test` (or` cd test && go test -v -failfast`)


chrisbsmith:
@moosecanswim can you provide a quick overview of how the test should flow? What is actually being tested, etc. would be good to add this as a comment to the release.go

moosecanswim:
also before each helm install the there is logic in place that determines weather or not to append `--set global.waiter.servicce_account.create=false` to the command.  This eliminates the need for the tester to determine if their install needs it.

moosecanswim:
@chrisbsmith i think i have a way to simplify adding new core services.  So if youre testing hold on a sec for a push.

chrisbsmith:
@moosecanswim it's highly possible that I'm just missing something, but where do we kick off tests? I don't see anywhere in your changes where `TestSetup` is called. You may have to treat me with kid gloves here.

moosecanswim:
Go tests will go through all tests in the current directory unless you
specify one with ‘—run <test bane>

chrisbsmith:
@moosecanswim got this warning when I ran `make secrets`

```
Makefile:161: warning: overriding commands for target `test'
helpers.mk:50: warning: ignoring old commands for target `test'
```

Makefile
```diff
@@ -156,3 +156,6 @@ lint-umbrella-charts:
 	ct lint --target-branch release-2.3 --chart-dirs .
```
chrisbsmith, line 156: Can you change this to `main`

```suggestion
	ct lint --target-branch main --chart-dirs .
```

chrisbsmith:
Also, this test failed (b/c of SLO and Postgres).

```
TestCoreReleases 2021-07-23T08:48:28-04:00 pod.go:98: Timedout waiting for the desired number of Pods to be created: 'Wait for num pods created to match desired count 3.' unsuccessful after 15 retries
    pod.go:67:
        	Error Trace:	pod.go:67
        	            				releases.go:170
        	            				greymatter_integration_test.go:89
        	Error:      	Received unexpected error:
        	            	'Wait for num pods created to match desired count 3.' unsuccessful after 15 retries
        	Test:       	TestCoreReleases
--- FAIL: TestCoreReleases (243.13s)
FAIL
exit status 1
FAIL	github.com/deciphernow/helm-charts/test	243.488s
make: *** [test] Error 1
➜  helm-charts git:(update-tes
```

However, the test did not tear down GM on failure. When this happens, I'd expect it to catch the failure and then back out everything it installed before tearing down k3d. Maybe this is what happens in CI but not when running `make test`?

chrisbsmith (approved):
LGTM

PR #968, opened 2021-07-23
clear charts dir efore dep up

moosecanswim:
clears out the charts dir before running a dep up


chrisbsmith (approved):
LGTM

moosecanswim:
ah cool... cause i merged it a min ago :-P

PR #972, opened 2021-07-28
Removes any specific reference to eks and uses kubernetes instead

chrisbsmith:

When testing against GKE, I found an issue where the redis pod was unable to write data due to permissions issues. Investigation showed that we were only applying a securityContext if we were in EKS. GKE also needed the securityContext for redis to be able to write data to disk.

In updating the code, I also realized that we were only checking if something was `eks` in a few places. I tested all of these places for `ne openshift` against EKS, GKE and K3d to verify that a value of `kubernetes` was sufficient.

So this PR removes the specific references to any environment setting of `eks`. Docs will also need to be updated.

none

Yes, all references to use `eks` as an environment have been removed.

It has been run against GKE, EKS and k3d.


jimgaspari (commented):
Changes look good.  Testing Deployment now.

moosecanswim (approved):
LGTM

PR #973, opened 2021-07-28
Helm docs

moosecanswim:
Update docs to point out only the most useful configurations.  This leverages helm-docs which is an opensource tool that replaces the gen docs python script.


chrisbsmith:
@moosecanswim is this something we can add to our github actions workflow? It's be awesome if we could automate the creation of this.

Or maybe we add it as a git commit hook?

moosecanswim:
@chrisbsmith Updated this so that it will build update the docs automatically if there is a change to a `values.yaml`, a readme template (`README.md.gotmpl`), `global.yaml`, or root `README.md`.

chrisbsmith:
@moosecanswim before merging, please squash these commits to clean the PR

Makefile
```diff
@@ -158,4 +158,16 @@ lint-umbrella-charts:
...
 
 test:
-	cd test && go test -v -failfast
+	- cd test && go test -v -failfast
+	- cd test && go test -v -run TestTearDown 
+
+update-global-configs:
+	- cd ci/helm-docs && ./global-configs.sh
+
+generate-configs:
```
chrisbsmith, line 167: Can we call this `generate-documentation` instead?

Makefile
```diff
@@ -158,4 +158,16 @@ lint-umbrella-charts:
 lint: lint-subcharts lint-edge-secrets lint-umbrella-charts
 
 test:
-	cd test && go test -v -failfast
+	- cd test && go test -v -failfast
+	- cd test && go test -v -run TestTearDown 
+
+update-global-configs:
```
chrisbsmith, line 164: Let's do a `s/configs/documentation/g` since we're actually updating documentation and not configurations.

ci/helm-docs/global-configs.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+
+HOME=$(git rev-parse --show-toplevel)
+
+DEST=fake-greymatter-chart
+
+#Can be yq if installed locally
```
chrisbsmith, line 7: I don't understand this comment.

chrisbsmith (changes requested):
All around this is a really good PR and I'm glad you were able to get it into the automated workflow. I'd like to see you change the name from `-configs` to `-documentation` because that's what the workflow is actually doing for us.

Makefile
moosecanswim: Sure.

ci/helm-docs/global-configs.sh
moosecanswim:
```
#Can be yq if installed locally
YQCMD="docker run -i docker.greymatter.io/internal/yq:2.4.1"
```
I choose to use a docker image to make sure the version someone has installed locally will not mess anything up.  So `YQCMD` can either be `yq` or this docker run statement.

moosecanswim:
@chrisbsmith  updated per comments

ci/helm-docs/global-configs.sh
chrisbsmith: OK, why don't you expand on the comment a little more so say ☝️ . For an external person, it may be a bit confusing

moosecanswim: Updated

chrisbsmith (dismissed):

chrisbsmith (approved):
LGTM

PR #975, opened 2021-08-03
increment spire charts for updated helm-docs release

moosecanswim:
the chart releaser for spire failed due to the charts already exisitng.  Here i am updating them.


chrisbsmith (approved):
lgtm

PR #976, opened 2021-08-03
update the path on push to include Chart.yaml because the global.yaml…

moosecanswim:
last update for helm-docs.  I forgot to include chart.yaml as a trigger for build docs.  The gloabal.yaml doc config info is based on a "fake chart" that pulls in data from fabric, sence, data etc charts.  This will make sure that top level requirements match the actually checked in charts when Chart.yaml are updated.


spire/README.md
```diff
@@ -2,7 +2,7 @@
 
 A Helm chart to deploy Spire
 
-![Version: 4.0.2](https://img.shields.io/badge/Version-4.0.2-informational?style=flat-square) ![AppVersion: 1.6.0](https://img.shields.io/badge/AppVersion-1.6.0-informational?style=flat-square)
+![Version: 4.0.4](https://img.shields.io/badge/Version-4.0.4-informational?style=flat-square) ![AppVersion: 1.6.0](https://img.shields.io/badge/AppVersion-1.6.0-informational?style=flat-square)
```
chrisbsmith, line 5: Is this something we'll have to change every time we update?

moosecanswim: That was automatically added in :-D

chrisbsmith: 👍

chrisbsmith (approved):
LGTM

PR #981, opened 2021-09-17
add new configuration for grey matter 1.7

tilleryd:
## Overview

This PR adds configuration for the gm-1.7 release. Primarily the following changes to support Health Checks:

- each services ingress listener needs to be configured with either a redis or nats metrics receiver (defaulting to redis, there are zero nats configs in this PR as it's another option)
- each service sidecar needs egress mesh configs to connect to redis
- new mesh configs for redis ingress
- new static sidecar config for control-api (upstream connection to redis)
- new static sidecar config for redis (downstream connection from control-api)

There are other small improvements and tweaks to files that I saw fit to make and are not directly related to the above, just general house keeping.

### Deploy with SPIRE
1. Clone this branch
2. Install the mesh
```
make k3d
make install
```
3. Generate user certs and navigate to the edge host `https://localhost:30000` to confirm everything is working

### Deploy with non-SPIRE TLS
1. Uninstall the mesh
```
make destroy-k3d
```
2. Disable SPIRE in global.yaml https://github.com/greymatter-io/helm-charts/blob/main/global.yaml#L76
3. Install the mesh
```
make k3d
make install
```
4. Generate user certs and navigate to the edge host `https://localhost:30000` to confirm everything is working


chrisbsmith (approved):
Everything launched as expected.  LGTM

dontlaugh:
Touch up the commit message when you squash :+1:

bcmendoza (changes requested):
I loaded the app and am not seeing health checks enabled. When I checked the Catalog API for health check results, I only saw the main `discovery_controlplane_reports` one. Is configuring health checks something we're leaving up to the user to do at a later time?

sense/values.yaml
```diff
@@ -216,7 +216,7 @@ catalog:
     # Name used for the deployment and service resources
     name: catalog
     # -- GM Catalog version
-    version: 2.0.0
+    version: 3.0.0-rc.1
```
bcmendoza, line 219:
```suggestion
    version: 3.0.0-rc.2
```

sense/catalog/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v2
-appVersion: 2.0.0
+appVersion: 3.0.0-rc.1
```
bcmendoza, line 2:
```suggestion
appVersion: 3.0.0-rc.2
```

sense/catalog/values.yaml
```diff
@@ -58,7 +58,7 @@ catalog:
   # Name used for the deployment and service resources
   name: catalog
   # GM Catalog version
-  version: 2.0.0
+  version: 3.0.0-rc.1
```
bcmendoza, line 61:
```suggestion
  version: 3.0.0-rc.2
```

sense/catalog/values.yaml
```diff
@@ -181,7 +181,7 @@ services:
     name: 'Grey Matter Catalog'
     owner: 'Decipher'
     data_prefix: "/data"
-    version: 2.0.0
+    version: 3.0.0-rc.1
```
bcmendoza, line 184:
```suggestion
    version: 3.0.0-rc.2
```

tilleryd:
@bcmendoza I think there may be an issue with the health feature flag for the dashboard. It's supposed to be enabled by default for the release image. 

We should enable all health checks rather than require customers to do so. This feature is a first class citizen of the product and one of the primary features of the release. With that said, how do I turn them on?

bcmendoza (approved):
These changes look good. As discussed, Catalog needs a new RC to handle retries for connecting to Redis. I totally overlooked that in developing this feature.

tilleryd:
@bcmendoza All good. Thanks for looking into it.

PR #982, opened 2021-09-22
add route redirects using regex

tilleryd:

This PR alters Grey Matter route configs. that are bootstrapped by Control API for core services. The deprecated "path" is removed and all redirects are done with a single config, rather than two, which is now accomplished via regex.

none

n/a

```
make k3d
make install
kc get secret greymatter-user-cert -o jsonpath="{.data['tls\.crt']}" | base64 -d > tls.crt
kc get secret greymatter-user-cert -o jsonpath="{.data['tls\.key']}" | base64 -d > tls.key
kc get secret greymatter-user-cert -o jsonpath="{.data['ca\.crt']}" | base64 -d > ca.crt
openssl pkcs12 -export -out greymatter.p12 -inkey tls.key -in tls.crt -certfile ca.crt -passout pass:password
```

Import greymatter.p12 and open https://localhost:30000 to ensure the dashboard renders. Click random service documentation links.


chrisbsmith:
@tilleryd you'll need to update the chart versions to pass the linting test.

fabric/control-api/config/bootstrap.sh
```diff
@@ -94,7 +94,7 @@ for d in */; do
         sleep $delay
     done
 
-    for file in route-*.json; do
+    for file in route.json; do
```
danielpcox, line 97: I notice this is a for loop over just one file

fabric/control-api/json/edge/route.json
```diff
@@ -0,0 +1,22 @@
...
+  "domain_key": "edge",
+  "zone_key": "{{.Values.global.zone}}",
+  "route_match": {
+    "path": "/services/{{.service.serviceName}}/latest/",
+    "match_type": "prefix"
+  },
+  "prefix_rewrite": "/",
+  "redirects": [
+    {
+      "from": "^/services/{{.service.serviceName}}/latest$",
```
danielpcox, line 12: One random thought: We're letting part of this regex come from the service name, so the service name will need to contain no regex special characters or else it _may_ behave unexpectedly. Though I guess this isn't a problem since we're the ones setting these serviceNames.

danielpcox (commented):
Looks good. My one comment isn't really an issue.

fabric/control-api/json/edge/route.json
tilleryd: A potential concern but I don't think this should block this PR. Can you approve?

danielpcox (approved):
Sorry, I meant to approve last time.

